PR #10, opened 2020-11-18
HMA-3786: New Menu Panel Component

ngoulongkam:
# 📝 Description

Github Issue
https://github.com/hmrc/android-components/issues/HMA-3786
  
- [x] Updated CHANGELOG
- [x] Updated README

# 🛠 Testing Notes
- Test new menu panel component

# 📸 Screenshots
image


sample/src/main/res/layout/fragment_menu_panel_row.xml
```diff
@@ -0,0 +1,92 @@
...
+            <uk.gov.hmrc.components.organism.menu.MenuPanelRowView
+                android:id="@+id/menu_example_2"
+                android:layout_width="match_parent"
+                android:layout_height="match_parent"
+                app:body="@string/menu_panel_example_2_body"
+                app:title="@string/menu_panel_example_2_title" />
+
+        </com.google.android.material.card.MaterialCardView>
+
+        <com.google.android.material.card.MaterialCardView
```
lojbrooks, line 73: Should the card view be a part of the component?

lojbrooks (changes requested):

components/src/main/java/uk/gov/hmrc/components/organism/menu/MenuPanelRowView.kt
```diff
@@ -0,0 +1,83 @@
...
+import uk.gov.hmrc.components.base.PaddedComponent
+import uk.gov.hmrc.components.databinding.ComponentMenuPanelRowBinding
+import uk.gov.hmrc.components.extensions.addRipple
+import uk.gov.hmrc.components.extensions.setMargins
+
+class MenuPanelRowView @JvmOverloads constructor(
+    context: Context,
+    attrs: AttributeSet? = null,
+    defStyleAttr: Int = 0
+) : RelativeLayout(context, attrs, defStyleAttr), PaddedComponent {
```
lojbrooks, line 33: Could this extend MaterialCardView rather than RelativeLayout? Then we wouldn't need it in the xml.

components/src/main/res/drawable/components_round_circle.xml
```diff
@@ -0,0 +1,10 @@
+<shape xmlns:android="http://schemas.android.com/apk/res/android"
```
lojbrooks, line 1: Can we name this to reflect that it is used for the `MenuPanelRow`? Same for the other one too.

README.md
```diff
@@ -654,6 +654,25 @@ If you need padding on your dividers, you can set `android:dividerPadding="@dime
...
+    android:layout_marginStart="@dimen/hmrc_spacing_16"
+    android:layout_marginEnd="@dimen/hmrc_spacing_16"
+    android:layout_marginBottom="@dimen/hmrc_spacing_8"
+    app:body="@string/menu_panel_placeholder_body"
+    app:title="@string/menu_panel_placeholder_title" />
+```
+
+Optionally, a choice of notification can be added to the right side of the title with `setNotification(body: String? = null)` method.
+Providing a body will show a notification with text inside a rounded square, otherwise it'll be a blank circle notification.
+The notification can also be remove by using `removeNotification()` method.
```
lojbrooks, line 674:
```suggestion
The notification can also be removed by using `removeNotification()` method.
```

lojbrooks (approved):

components/src/main/java/uk/gov/hmrc/components/organism/menu/MenuPanelRowView.kt
```diff
@@ -0,0 +1,138 @@
...
+        setDefaultContentDescription(getTitle(), getBody())
+    }
+
+    override fun removeChildPadding() {
+        val defaultPadding = resources.getDimensionPixelSize(R.dimen.hmrc_spacing_24)
+        val leftRightPadding = resources.getDimensionPixelSize(R.dimen.hmrc_spacing_16)
+
+        binding.textTitle.setMargins(leftRightPadding, defaultPadding, leftRightPadding, defaultPadding)
+        binding.textBody.setMargins(leftRightPadding, 0, leftRightPadding, defaultPadding)
+    }
```
jemmaSlater, line 133: I don't think we need to have this logic here. We can just set the margins we need in the xml. This class does not really need to extend `PaddedComponent`. 

We use "PaddedComponent" for components we have built to allow children to be added when used (eg. adding a button or some extra text to a PrimaryCardView.). Different components have different padding rules, so eg. if we added a PrimaryButton to the PrimaryCardView, we would want to keep the Card component's padding, but if we added a SecondaryButton we would want to remove the child padding so that the touch state would be flush with the edges of the card.

We are not allowing children to be added to this component, so we do not need this logic. :)

components/src/main/res/layout/component_menu_panel_row.xml
```diff
@@ -0,0 +1,92 @@
...
+            android:layout_height="wrap_content"
+            android:importantForAccessibility="no"
+            android:gravity="start"
+            android:paddingStart="@dimen/hmrc_spacing_16"
+            android:paddingEnd="@dimen/hmrc_spacing_16"
+            android:paddingBottom="@dimen/hmrc_spacing_24" />
+
+    </LinearLayout>
+
+        <ImageView
```
jemmaSlater, line 79: The indentation here isn't quite right. Could do with a quick "Auto-Indent lines" :)

sample/src/main/java/uk/gov/hmrc/components/sample/organisms/MenuPanelRowFragment.kt
```diff
@@ -0,0 +1,70 @@
...
+
+class MenuPanelRowFragment : BaseComponentsFragment() {
+
+    private var binding: FragmentMenuPanelRowBinding by autoCleared()
+
+    override fun provideToolbar() = ToolbarState(true, R.string.organisms_menu_panel_row, true)
+
+    override fun onCreateView(inflater: LayoutInflater, container: ViewGroup?, savedInstanceState: Bundle?): View? {
+        binding = FragmentMenuPanelRowBinding.inflate(inflater, container, false)
+        binding.root.setBackgroundColor(ContextCompat.getColor(context!!, R.color.hmrc_white_background))
```
jemmaSlater, line 38: Rather than set this here, I think it would be better to just set it in the xml.

sample/src/main/res/layout/fragment_menu_panel_row.xml
```diff
@@ -0,0 +1,90 @@
...
+
+     Unless required by applicable law or agreed to in writing, software
+     distributed under the License is distributed on an "AS IS" BASIS,
+     WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+     See the License for the specific language governing permissions and
+     limitations under the License.
+-->
+<ScrollView xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:app="http://schemas.android.com/apk/res-auto"
+    style="@style/SampleScrollView">
```
jemmaSlater, line 19: You don't need to use this style here. It just contains the background colour which you are currently changing anyway. :) You can just set the colour we want here without using the style.

sample/src/main/res/layout/fragment_menu_panel_row.xml
```diff
@@ -0,0 +1,90 @@
+<?xml version="1.0" encoding="utf-8"?>
```
jemmaSlater, line 1: This file could also do with a quick "Auto-Indent lines"

sample/src/main/res/values/strings.xml
```diff
@@ -345,4 +346,18 @@
...
+    <string name="menu_panel_example_1_title">Pay As You Earn (PAYE)</string>
+    <string name="menu_panel_example_1_body">Check your tax codes and Income Tax from PAYE income sources</string>
+    <string name="menu_panel_example_2_title">Messages</string>
+    <string name="menu_panel_example_2_body">Messages and letters from HMRC</string>
+    <string name="menu_panel_example_3_title">Help to Save account</string>
+    <string name="menu_panel_example_3_body">A savings account with bonuses designed to help you start saving</string>
+    <string name="menu_panel_example_4_title">News</string>
+    <string name="menu_panel_example_4_body">News and current tax related information from HMRC</string>
+    <string name="menu_panel_example_5_title">Long Long Long Long Long Long Long Long Long Long Long Long Long Long Long Long Title</string>
+    <string name="menu_panel_example_5_body">Really Really Really Really Really Really Really Really Really Long Body</string>
```
jemmaSlater, line 361: Instead of defining these to provide long text examples, you can use our pre-existing "long_text" and "longer_text" strings that are at the top of this file.

jemmaSlater (changes requested):

jemmaSlater (approved):

jemmaSlater (approved):

PR #19, opened 2021-01-21
Info message duplicated buttons fix

jemmaSlater:
# 📝 Description

Ensure buttons are cleared before adding new ones to prevent duplicate buttons.
  
- [x] Updated CHANGELOG
- [x] Updated README

# 🛠 Testing Notes

# 📸 Screenshots

### DIFF
--- a/components/src/main/java/uk/gov/hmrc/components/organism/information/InformationMessageCardView.kt
+++ b/components/src/main/java/uk/gov/hmrc/components/organism/information/InformationMessageCardView.kt
@@ -91,6 +91,8 @@ class InformationMessageCardView @JvmOverloads constructor(
     }
 
     fun addHeadlineButtons(buttons: List<SecondaryButton>) {
+        binding.headlineButtonsContainer.removeAllViews()
+
         val spacing8 = context.resources.getDimensionPixelSize(R.dimen.hmrc_spacing_8)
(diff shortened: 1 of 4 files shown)


chrisob55 (approved):

components/src/main/java/uk/gov/hmrc/components/organism/information/InformationMessageCardView.kt
```diff
@@ -91,6 +91,8 @@ class InformationMessageCardView @JvmOverloads constructor(
     }
 
     fun addHeadlineButtons(buttons: List<SecondaryButton>) {
+        binding.headlineButtonsContainer.removeAllViews()
```
lojbrooks, line 94: Can we rename to `setHeadlineButtons` if we need to do this?

components/src/main/java/uk/gov/hmrc/components/organism/information/InformationMessageCardView.kt
```diff
@@ -105,6 +107,8 @@ class InformationMessageCardView @JvmOverloads constructor(
     }
 
     fun addContentButtons(buttons: List<Button>) {
+        binding.buttonsContainer.removeAllViews()
```
lojbrooks, line 110: Same here, `setContentButtons`

lojbrooks (changes requested):

PR #23, opened 2021-02-02
Bugfix: info messages update content visibility

jemmaSlater:
# 📝 Description

In the future we will update InformationMessageCardViews to have a new content area within the coloured panel. For now, we just remove the content area (title, body, buttons) below the coloured panel to ensure we are no longer able to show text in the old-style design format.
  
- [x] Updated CHANGELOG
- [x] Updated README

# 🛠 Testing Notes

# 📸 Screenshots


components/src/main/java/uk/gov/hmrc/components/organism/information/InformationMessageCardView.kt
```diff
@@ -80,14 +80,24 @@ class InformationMessageCardView @JvmOverloads constructor(
...
+            binding.viewBottomPadding.visibility = View.VISIBLE
+        }
+        updateContentContainerVisibility()
     }
 
-    fun setBody(body: CharSequence?) {
+    fun setContentBody(body: CharSequence?) {
         binding.contentBody.text = body
-        binding.contentBody.visibility = if (body != null) View.VISIBLE else View.GONE
+        binding.contentBody.visibility = if (!body.isNullOrEmpty()) View.VISIBLE else View.GONE
```
chrisob55, line 96: Bit of mind bend with "if not is null or is not empty"

Might it read better like?

```kotlin
binding.contentBody.visibility = if (body.isNullOrEmpty()) View.GONE else View.VISIBLE
```

chrisob55: But not precious about it if you disagree

chrisob55 (changes requested):

components/src/main/java/uk/gov/hmrc/components/organism/information/InformationMessageCardView.kt
jemmaSlater: This change would definitely have made sense! However, just deleted it now as the requirements have changed. :)

chrisob55 (approved):

PR #38, opened 2021-05-05
TextInputView accessibility improvements

jemmaSlater:
# 📝 Description

When talkback focus is given to a TextInputView, we want the user to be made aware of any errors or character limits associated with that view. We also want to be able to support custom hint content descriptions.
  
- [-] Updated CHANGELOG

# 🛠 Testing Notes

When talkback focus is given to a TextInputView, it should announce "[hint], [character limit], [error]" as required. 
Note: The second example ("Message") on the Text Input View screen in the sample app already contains a custom hint message. 
Also need to check that the Currency Input View is working as expected.

# 📸 Screenshots

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,9 +16,13 @@ Allowed headings:
 
 ## [Unreleased]
 
-### Changed
+### Added
 
 * Added ability to set error content descriptions on `TextInputView`
+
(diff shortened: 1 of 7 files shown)


chrisob55 (approved):
LGTM from a code perspective - this one will be all about the QA

PR #41, opened 2021-05-14
Add click listeners to multicolumnrow texts

jemmaSlater:
# 📝 Description

Adds the ability to add a click listener, and custom click content description, to a column in the MultiColumnRowView. For example, this could be used to enable 'press to copy' functionality on a particular column in a row.
  
- [x] Updated CHANGELOG
- [x] Updated README

# 🛠 Testing Notes
I have adapted the Sample app so that the placeholder MultiColumnRowView has the following configuration: 
Text1 = no click listener (same behaviour as before)
Text2 = click listener to show toast, no custom click content description
Text3 = click listener to show toast, custom click content description

# 📸 Screenshots


components/src/main/java/uk/gov/hmrc/components/molecule/item/MultiColumnRowView.kt
```diff
@@ -173,6 +174,27 @@ class MultiColumnRowView @JvmOverloads constructor(
         }
     }
 
+    fun setText1ClickAction(listener: OnClickListener, clickDescription: CharSequence? = null) {
```
JamieCruwys, line 177: Can we switch the order around of these so the listener is at the end? This will allow us do do things like this in Kotlin:
`setText1ClickAction { Toast.makeText(context, "text 1 custom action invoked", LENGTH_LONG).show() }`

JamieCruwys (changes requested):

components/src/main/java/uk/gov/hmrc/components/molecule/item/MultiColumnRowView.kt
jemmaSlater: Good suggestion! Changed in https://github.com/hmrc/android-components/pull/41/commits/69a87ddd2cebf6fcbab6af997ae05d35b2fcbd17 :)

JamieCruwys (approved):

ngoulongkam (approved):

PR #63, opened 2021-11-04
HMA-5118 Update status view to support vectors wider than 100dp

JamieCruwys:
# 📝 Description

The pay by bank vector in https://jira.tools.tax.service.gov.uk/browse/HMA-5118 is `200dp` wide.

`StatusView` fixes the width of the icon to be `100dp`. This means that the vector is scaled down.

To fix this I have set `minWidth` to `100dp` and the `width` to `wrap_content`.

This will mean that for all the existing icons, it'll still be 100dp wide, but for icons wider than 100dp, it will take up more horizontal space on the screen. This is fine because in `StatusView` there is nothing on either side of the icon, so it can take up all the space it wants.

- [x] Updated CHANGELOG
- [x] Updated README (not needed)

# 🛠 Testing Notes

Retest `StatusView` and `StatusCard` usages

# 📸 Screenshots

FYI - the vector was temporarily added to generate these screenshots

| Before | After |
| ------ | ------ |
| HMA-5118-components-update-before | HMA-5118-components-update |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@ Allowed headings:
 
 ## [Unreleased]
 
-* Updated `StatusCardView` to support vectors than are more than 100dp wide
+* Updated `StatusCardView` to improve support for vectors than are more than 100dp wide
 
 ## [3.18.0] - 2021-10-22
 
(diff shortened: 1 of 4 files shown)


chrisob55 (approved):

chrisob55 (approved):

chrisob55 (approved):

JamieCruwys:
QA'd by Will

PR #70, opened 2022-01-24
Add ability to remove default padding from WarningView

jemmaSlater:
# 📝 Description

The WarningView comes with a default padding of 8dp. In some cases, this makes the layout too spaced out (eg. when nested inside a `HeadlineCardView`). To minimise changes to the app, this adds an attribute where we can remove this default padding if needed, meaning existing usages will remain visually unaffected.
  
- [x] Updated CHANGELOG
- [ ] Updated README

# 🛠 Testing Notes

# 📸 Screenshots
Screenshot 2022-01-24 at 17 54 03


JamieCruwys:
Would it be possible to not use a negative for this and instead use something like `withPadding` or something with the default value being true? The double negatives can make it harder to read

chrisob55 (dismissed):

jemmaSlater:
> Would it be possible to not use a negative for this and instead use something like `withPadding` or something with the default value being true? The double negatives can make it harder to read

Sure! Was a little conscious of it sounding like you could provide your own padding value, so have just gone for `defaultPadding` to have behaviour in line with our existing `childPadding` which also defaults to true.

CHANGELOG.md
```diff
@@ -16,6 +16,10 @@ Allowed headings:
 
 ## [Unreleased]
 
+### Added
+
+* `WarningView` has new attribute `defaultPadding` which removes the default padding from the component when set to false. When not set, component has `8dp` padding.
```
JamieCruwys, line 21:
```suggestion
* `WarningView` has new attribute `defaultPadding` which removes the default padding from the component when set to false. When this is true, component has `8dp` padding.
```

JamieCruwys (changes requested):

CHANGELOG.md
JamieCruwys: ^ or similar. Basically the statement isn't correct anymore so needs updating

jemmaSlater: Have updated to clarify - but we do not need to set this explicitly to enable the default padding, it only needs to be set in the xml if we want to remove the padding. (This is a bit against the way we usually work - define if "true" - but this is to prevent affecting any existing usages of the component)

JamieCruwys (approved):

PR #74, opened 2022-03-23
HMA-5565: Update InformationMessageCardView to show outline style button

ngoulongkam:
# 📝 Description

Github Issue
  
- [/] Updated CHANGELOG
- [/] Updated README

# 🛠 Testing Notes

# 📸 Screenshots
image


components/src/main/java/uk/gov/hmrc/components/organism/information/InformationMessageCardView.kt
```diff
@@ -92,26 +92,45 @@ class InformationMessageCardView @JvmOverloads constructor(
...
+                    setTextColor(ContextCompat.getColorStateList(context, type!!.headlineTint))
+                }
+            } else {
+                it.button.backgroundTintList = ContextCompat.getColorStateList(context, R.color.hmrc_white)
+            }
+            binding.headlineButtonsContainer.addView(it.button)
         }
     }
 
+    data class InformationMessageButton(
```
JamieCruwys, line 121: It might be easier for callers if this was put into its own class. At the moment they will need to write `InformationMessageCardView.InformationMessageButton()`. If you extract it out then they can call `InformationMessageButton()` directly. Of course they could import the longer version but extracting this data class would make it a bit tidier.

JamieCruwys (changes requested):
Would be a bit cleaner for the library consumer if the data class was extracted.
Great work though @ngoulongkam

components/src/main/java/uk/gov/hmrc/components/organism/information/InformationMessageCardView.kt
jemmaSlater, line 121: Do you think this would be tidier as a sealed class? Then rather than having to set true/false, you just define the one you want.

```
sealed class InformationMessageButton(val button: SecondaryButton, val isOutlineButton: Boolean {
  data class ActionButton(actionButton: Secondary button): InformationMessageButton(actionButton, false)
  data class OutlineButton(outlineButton: SecondaryButton): InformationMessageButton(outlineButton, true)
```

jemmaSlater (approved):

PR #84, opened 2022-11-21
Feature/HMA-6225-donut-chart-view

emmaatkins:
# 📝 Description

Github Issue
https://github.com/hmrc/android-components/issues/HMA-6225
  
- [x] Updated CHANGELOG
- [  ] Updated README

# 🛠 Testing Notes
Check new component - donut chart view + examples

# 📸 Screenshots
donut


CHANGELOG.md
```diff
@@ -16,6 +16,12 @@ Allowed headings:
 
 ## [Unreleased]
 
+## [4.3.0] - 2022-11-21
```
jemmaSlater, line 19: I don't think we need to add the version/date to the changelog a this stage. In theory there might be a few changes per release, so I _think_ the release process adds this bit for us when we actually do the release with the date at that time.

components/src/main/res/values/colors.xml
```diff
@@ -48,4 +48,7 @@
     <color name="hmrc_information_message_urgent_headline_background">@color/hmrc_red</color>
     <color name="hmrc_information_message_notice_headline_background">@color/hmrc_black</color>
 
+    <color name="donut_chart_color_1">@color/hmrc_teal</color>
+    <color name="donut_chart_color_2">#003078</color>
```
jemmaSlater, line 52: Just for consistency with the other colours provided by the component library, I think we should name these with the "hmrc" prefix too. (remember to also update the dark mode colours, and the publicly exposed names!) 
```suggestion
    <color name="hmrc_donut_chart_color_1">@color/hmrc_teal</color>
    <color name="hmrc_donut_chart_color_2">#003078</color>
```

components/src/main/res/values/public.xml
```diff
@@ -25,6 +25,8 @@
     <public name="hmrc_information_message_urgent_headline_background" />
     <public name="hmrc_information_message_notice_headline_background" />
     <public name="hmrc_secondary_button_ripple" type="color" />
+    <public name="hmrc_donut_1" type="color" />
+    <public name="hmrc_donut_2" type="color" />
```
jemmaSlater, line 29: I think these names need to match the ones in the colors.xml file.

PR #86, opened 2022-12-14
HMA-6127:new_component_EditableListView_to_library

sajinabraham:
# 📝 Description
https://jira.tools.tax.service.gov.uk/browse/HMA-6127
Github Issue
https://github.com/hmrc/android-components
  
- [ ] Updated CHANGELOG
- [ ] Updated README

# 🛠 Testing Notes

# 📸 Screenshots


components/src/main/java/uk/gov/hmrc/components/organism/editable/EditableListView.kt
```diff
@@ -0,0 +1,81 @@
...
+    lateinit var editableListViewAdapter: EditableListViewAdapter
+    var editableItem = ArrayList<EditableItem>()
+    var itemPosition: Int? = null
+    var clicked = false
+
+    init {
+
+        binding.secondaryButton.apply {
+
+            setCompoundDrawablesWithIntrinsicBounds(R.drawable.ic_edit, 0, 0, 0)
```
jemmaSlater, line 45: If we use the `IconButton` component rather than `SecondaryButton` then we don't need to add the icon in this way. However, I think it does still need `setIconResource(icon)` (see `IconButtonCardView` for an example of this)

components/src/main/java/uk/gov/hmrc/components/organism/editable/EditableListView.kt
```diff
@@ -0,0 +1,81 @@
...
+
+            setOnClickListener {
+                editableListViewAdapter.isEditEnable = !editableListViewAdapter.isEditEnable
+                if (clicked) {
+                    clicked = false
+                    announceForAccessibility("Edit buttons now hidden")
+
+                    setCompoundDrawablesWithIntrinsicBounds(R.drawable.ic_edit, 0, 0, 0)
+                    text = context.getString(R.string.edit_button_update_or_remove)
+                    setAccessibilityMessage("edit your company benefits")
```
jemmaSlater, line 56: We need to make the component so it can be used in various places for different purposes. So we don't want to include anything that is specific to the "your employment" screen use case in the component itself. We should instead build it in a way that allows the user to provide all the strings/icons/content descriptions. See `MenuPanelRowView` for an example. We should support both programmatically constructed and xml attributes.

components/src/main/res/layout/component_editable_list_view.xml
```diff
@@ -0,0 +1,38 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<androidx.constraintlayout.widget.ConstraintLayout xmlns:android="http://schemas.android.com/apk/res/android"
```
jemmaSlater, line 3: It doesn't look like we are making use of any of the constraint layout features in this layout? If we don't need it we can keep it simple and just use a LinearLayout.

components/src/main/res/layout/component_editable_list_view.xml
```diff
@@ -0,0 +1,38 @@
...
+            android:layout_height="wrap_content"
+            app:layoutManager="androidx.recyclerview.widget.LinearLayoutManager"
+            tools:listitem="@layout/editable_list_item" />
+
+        <uk.gov.hmrc.components.atom.button.SecondaryButton
+            android:id="@+id/secondary_button"
+            style="@style/Text.Body"
+            android:layout_width="wrap_content"
+            android:layout_height="wrap_content"
+            android:text="@string/accessibility_counter_limit_exceeded" />
```
jemmaSlater, line 35: I don't think we want to define this here?

components/src/main/res/layout/component_editable_list_view.xml
```diff
@@ -0,0 +1,38 @@
...
+        <androidx.recyclerview.widget.RecyclerView
+            android:id="@+id/list_item"
+            android:layout_width="match_parent"
+            android:layout_height="wrap_content"
+            app:layoutManager="androidx.recyclerview.widget.LinearLayoutManager"
+            tools:listitem="@layout/editable_list_item" />
+
+        <uk.gov.hmrc.components.atom.button.SecondaryButton
+            android:id="@+id/secondary_button"
+            style="@style/Text.Body"
```
jemmaSlater, line 32: If we change this to use the IconButton we shouldn't need to define a style here, as we want the standard usage of our existing IconButton atom.

components/src/main/res/layout/editable_list_item.xml
```diff
@@ -0,0 +1,31 @@
...
+    xmlns:app="http://schemas.android.com/apk/res-auto"
+    xmlns:tools="http://schemas.android.com/tools"
+    android:id="@+id/motionLayout"
+    android:layout_width="match_parent"
+    android:layout_height="32dp"
+    app:layoutDescription="@xml/scene_editing_mode"
+    tools:showPaths="true">
+
+    <TextView
+        android:id="@+id/benefit_name"
```
jemmaSlater, line 12: can we name this something more generic? We can be consistent with the design invision file and call it `column_one`

components/src/main/res/layout/editable_list_item.xml
```diff
@@ -0,0 +1,31 @@
...
+        style="@style/Text.Body"
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        android:gravity="start" />
+
+    <TextView
+        android:id="@+id/benefitAmount"
```
jemmaSlater, line 19: and then this one can be `column_two`

components/src/main/res/layout/editable_list_item.xml
```diff
@@ -0,0 +1,31 @@
...
+        style="@style/Text.Body"
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        android:gravity="center_vertical|end" />
+
+    <uk.gov.hmrc.components.atom.button.SecondaryButton
+        android:id="@+id/editButton"
+        android:text="@string/edit"
```
jemmaSlater, line 27: Can we build it so the consumer of the component provides this text? It will make things like translation easier. :)

components/src/main/res/values/strings.xml
```diff
@@ -41,4 +41,9 @@
 
     <string name="currency_input_prefix">£</string>
     <string name="text_input_clear_button_content_description">Clear text</string>
+
+    <!-- Edit Button Text -->
+    <string name="edit">Edit</string>
+    <string name="edit_button_update_or_remove">Update or remove benefits</string>
+    <string name="edit_button_finish">Finish updating benefits</string>
```
jemmaSlater, line 48: So with the above changes we should be able to remove these.. we want to be able to provide all strings through the app that pulls it in.

components/src/main/res/xml/scene_editing_mode.xml
```diff
@@ -0,0 +1,87 @@
+<?xml version="1.0" encoding="utf-8"?>
+<MotionScene xmlns:android="http://schemas.android.com/apk/res/android"
```
jemmaSlater, line 2: Nice! I think this is the first opportunity we've had to use MotionScene in our app! :D

components/src/main/res/xml/scene_editing_mode.xml
```diff
@@ -0,0 +1,87 @@
...
+
+    </Transition>
+
+    <ConstraintSet android:id="@+id/start">
+
+        <Constraint
+            android:id="@+id/benefit_name"
+            android:layout_width="0dp"
+            android:layout_height="wrap_content"
+            android:layout_marginStart="8dp"
```
jemmaSlater, line 32: for margin sizes please can we use our pre-defined dimens?  So in this case you can use `hmrc_spacing_8`

components/src/main/res/xml/scene_editing_mode.xml
```diff
@@ -0,0 +1,87 @@
...
+            motion:layout_constraintBottom_toBottomOf="parent"
+            motion:layout_constraintEnd_toEndOf="parent"
+            motion:layout_constraintStart_toEndOf="@id/benefit_name"
+            motion:layout_constraintTop_toTopOf="parent" />
+
+        <Constraint
+            android:id="@+id/editButton"
+            android:layout_width="40dp"
+            android:layout_height="40dp"
```
jemmaSlater, line 50: What do these define? If it is something the user is meant to be able to touch this isn't quite big enough for the touch state to pass our accessibility requirements, which should be `48dp`. Or is it the button that contains the word "edit", in which case, can we not use wrap_content? How will this work with large text?

sample/src/main/java/uk/gov/hmrc/components/sample/organisms/EditableListViewFragment.kt
```diff
@@ -0,0 +1,58 @@
...
+        savedInstanceState: Bundle?
+    ): View {
+        binding = FragmentEditableListViewBinding.inflate(inflater, container, false)
+
+        return binding.root
+    }
+
+    override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
+        super.onViewCreated(view, savedInstanceState)
+        setHasOptionsMenu(true)
```
jemmaSlater, line 49: Do we need this?

sample/src/main/res/drawable/ic_edit.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
```
jemmaSlater, line 2: We shouldn't need to add this to both the component module and the sample module. We either need to build it in to the component (which I think is fine for this one) or have it supplied by the consumer. So I think we can remove this one, as we have added it to the component itself.

sample/src/main/res/drawable/ic_tick.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
```
jemmaSlater, line 2: We can remove this one too.

sample/src/main/res/layout/fragment_editable_list_view.xml
```diff
@@ -0,0 +1,29 @@
...
+
+<ScrollView xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:app="http://schemas.android.com/apk/res-auto"
+    style="@style/SampleScrollView">
+
+    <LinearLayout
+        android:id="@+id/content_layout"
+        style="@style/SampleSectionContainer">
+
+        <uk.gov.hmrc.components.atom.heading.Heading5 style="@style/PlaceholderText" />
```
jemmaSlater, line 11: I think we can simplify and for consistency, just use a text view here.
```suggestion
        <TextView style="@style/PlaceholderText" />
```

sample/src/main/res/layout/fragment_editable_list_view.xml
```diff
@@ -0,0 +1,29 @@
...
+
+        <uk.gov.hmrc.components.organism.primary.PrimaryCardView
+            android:id="@+id/headline_placeholder"
+            android:layout_width="match_parent"
+            android:layout_height="wrap_content"
+            android:layout_margin="@dimen/hmrc_spacing_16"
+            app:headline="@string/headline_placeholder_headline"
+            app:title="Your Company Benefits">
```
jemmaSlater, line 19: The placeholder should just be the basic bare boned component, how it comes, out of the box.. so we shouldn't put it on a card cause the card doesn't come built in to the component. The examples section is for that!

Because this component doesn't have a background colour built in, we can use `<FrameLayout style="@style/PlaceholderContainer">` to give it a temporary background for this placeholder example.

jemmaSlater (changes requested):

components/src/main/res/values/attrs.xml
```diff
@@ -253,4 +253,15 @@
...
 
+    <declare-styleable name="EditableListView">
+        <!-- Title to display. -->
+        <attr name="title" />
+        <!-- button Title to display. -->
+        <attr name="buttontitle" />
+        <!-- Button Text to display. -->
+        <attr name="editButtonText" />
+        <!-- Icon to display. -->
+        <attr name="icon" />
```
jemmaSlater, line 264: I don't think you're using all of these? 
I think we do need to provide the "title", "button start editing text", "button finish editing text" and the "edit button text". We'll probably also need content description text for the edit button, as I think the first implementation of it has an "opens in your web browser" accessibility requirement.

components/src/main/res/layout/component_editable_list_view.xml
```diff
@@ -0,0 +1,33 @@
+<?xml version="1.0" encoding="utf-8"?>
+<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:app="http://schemas.android.com/apk/res-auto"
+    xmlns:tools="http://schemas.android.com/tools"
+    android:layout_width="match_parent"
+    android:layout_height="match_parent"
+    android:layout_marginStart="@dimen/hmrc_spacing_16"
+    android:layout_marginEnd="@dimen/hmrc_spacing_16"
```
jemmaSlater, line 8: I don't think we want these margins.
One reason is because usually we want the ripple of the secondary/icon buttons to stretch to the sides of the view, and so having the margin on the parent layout cuts into the ripple. The button has built in margins that mean the content is already spaced properly from the sides.
But also, this component will be added to an existing card, so if we don't have any padding it leaves it more flexible to fit into other layouts, which can add the padding as needed.

components/src/main/res/layout/editable_list_item.xml
```diff
@@ -0,0 +1,33 @@
...
+
+    <TextView
+        android:id="@+id/column_one"
+        style="@style/Text.Body"
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        android:gravity="start" />
+
+    <TextView
+        android:id="@+id/column_two"
```
jemmaSlater, line 19: Can we give this one a bit of space between the columns? I think probably 16dp (remember to use our hmrc dimen)
On my emulator they get a bit squashed together when the edit button is showing and the text is long.

jemmaSlater (changes requested):

jemmaSlater (changes requested):

sample/src/main/res/layout/fragment_editable_list_view.xml
```diff
@@ -0,0 +1,51 @@
...
+                android:layout_width="match_parent"
+                android:layout_height="match_parent" />
+        </FrameLayout>
+
+        <TextView style="@style/ExampleText" />
+
+        <com.google.android.material.card.MaterialCardView
+            android:layout_width="match_parent"
+            android:layout_height="wrap_content"
+            android:layout_marginStart="@dimen/hmrc_spacing_16"
```
jemmaSlater, line 25: Can we add `android:paddingTop="@dimen/hmrc_spacing_16` to these example cards to make them look a bit more like they will when used in the app?

components/src/main/java/uk/gov/hmrc/components/organism/editable/EditableListView.kt
```diff
@@ -148,23 +143,25 @@ open class EditableListView @JvmOverloads constructor(
...
-                        } else {
+                    if (event.eventType == AccessibilityEvent.TYPE_VIEW_ACCESSIBILITY_FOCUSED)
+                        if (child?.id == binding.title.id) {
                             binding.iconButton.accessibilityTraversalBefore = nextFocusForwardId
+                        } else if (child?.id == binding.iconButton.id) {
+                            if (editMode) {
+                                binding.iconButton.accessibilityTraversalBefore = binding.title.id
+                            } else {
+                                binding.iconButton.accessibilityTraversalBefore = nextFocusForwardId
+                            }
```
emmaatkins, line 163: The if and else are the same - these could be combined?

jemmaSlater (changes requested):

components/src/main/java/uk/gov/hmrc/components/organism/editable/EditableListView.kt
```diff
@@ -0,0 +1,174 @@
...
+import androidx.recyclerview.widget.DividerItemDecoration
+import androidx.recyclerview.widget.LinearLayoutManager
+import uk.gov.hmrc.components.R
+import uk.gov.hmrc.components.databinding.ComponentEditableListViewBinding
+import kotlin.random.Random
+
+open class EditableListView @JvmOverloads constructor(
+    context: Context,
+    attrs: AttributeSet? = null,
+) : ConstraintLayout(context, attrs) {
```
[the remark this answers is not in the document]
jemmaSlater: Please can we add the additional style parameters incase we need them?

components/src/main/java/uk/gov/hmrc/components/organism/editable/EditableListView.kt
```diff
@@ -0,0 +1,174 @@
...
+    }
+
+    fun setButtonIconData(
+        @DrawableRes startEditingIcon: Int,
+        @DrawableRes endEditingIcon: Int
+    ) {
+        buttonIcon = Pair(startEditingIcon, endEditingIcon)
+        binding.iconButton.apply {
+            setIconResource(if (editMode) buttonIcon.second else buttonIcon.first)
+        }
```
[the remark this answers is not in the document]
jemmaSlater: Is there a reason we can't use `endEditingIcon` and `startEditingIcon` directly? I don't think we gain anything by creating the pair..

components/src/main/java/uk/gov/hmrc/components/organism/editable/EditableListView.kt
```diff
@@ -0,0 +1,174 @@
...
+    }
+
+    fun setButtonAccessibility(
+        startEditingAccessibility: String,
+        endEditingAccessibility: String
+    ) {
+        buttonAccessibility = Pair(startEditingAccessibility, endEditingAccessibility)
+        binding.iconButton.apply {
+            announceForAccessibility(if (editMode) buttonAccessibility.second else buttonAccessibility.first)
+        }
```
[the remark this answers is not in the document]
jemmaSlater: same as above still here

components/src/main/java/uk/gov/hmrc/components/organism/editable/EditableListViewAdapter.kt
```diff
@@ -0,0 +1,73 @@
...
+            columnOne.text = result.name
+            columnTwo.text = result.value
+            iconButton.text = result.buttonText
+            itemView.contentDescription =
+                if (isEditEnable) {
+                    "${result.name}: ${result.valueContentDescription} Tap to ${result.buttonText} ${result.name} " +
+                        "Item $position of $itemCount."
+                } else {
+                    "${result.name}: ${result.valueContentDescription} Item $position of $itemCount."
+                }
```
jemmaSlater, line 61: You're still missing some full stops. And I think we can remove the name part.. so it will just announce eg.  "Medical: £500. Tap to edit."

```suggestion
                    "${result.name}: ${result.valueContentDescription}. Tap to ${result.buttonText}." +
                        "Item $position of $itemCount."
                } else {
                    "${result.name}: ${result.valueContentDescription}. Item $position of $itemCount."
                }
```

jemmaSlater (approved):

jemmaSlater (approved):

components/src/main/java/uk/gov/hmrc/components/organism/editable/EditableListViewAdapter.kt
```diff
@@ -0,0 +1,77 @@
...
+            columnTwo.text = result.value
+            iconButton.text = result.buttonText
+            val positionLabel = position + 1
+            divider.isVisible = positionLabel < itemCount
+            itemView.contentDescription =
+                if (isEditEnable) {
+                    "${result.name}: ${result.valueContentDescription}. Tap to ${result.buttonText}." +
+                        "Item $positionLabel of $itemCount."
+                } else {
+                    "${result.name}: ${result.valueContentDescription}. Item $positionLabel of $itemCount."
```
emmaatkins, line 64: Space after the full stop please and change the full stop to a comma.

Please check other locations where using a full stop to see if a comma would be more consistent.

emmaatkins (changes requested):

emmaatkins (approved):

components/src/main/java/uk/gov/hmrc/components/organism/editable/EditableListViewAdapter.kt
```diff
@@ -0,0 +1,81 @@
...
+            }
+
+            if (isEditEnable) {
+                motionLayout.transitionToEnd()
+            } else {
+                motionLayout.transitionToStart()
+            }
+        }
+
+        fun valueAccessibility(editableItem: EditableListView.EditableItem): String {
```
jemmaSlater, line 77: This can be a private function

jemmaSlater (approved):

PR #102, opened 2023-02-14
Hma 5068 buttons(atom)

sajinabraham:
# 📝 Description

Github Issue
https://jira.tools.tax.service.gov.uk/browse/HMA-5068
  
- [ ] Updated CHANGELOG
- [ ] Updated README

# 🛠 Testing Notes

# 📸 Screenshots


components-compose/src/main/java/uk/gov/hmrc/components/compose/atom/button/Button.kt
```diff
@@ -15,4 +15,119 @@
...
+    textAlign: TextAlign = TextAlign.Center,
+    enabled: Boolean,
+    shape: Shape = RectangleShape,
+    buttonColors: ButtonColors = ButtonDefaults.buttonColors(),
+    contentPadding: PaddingValues = ButtonDefaults.ContentPadding,
+    onClick: () -> Unit,
+    content: @Composable RowScope.() -> Unit = {},
+) {
+    Button(
+        modifier = modifier.heightIn(min = hmrc_button_size_48).fillMaxWidth(), colors = buttonColors,
```
ngoulongkam, line 61: The current Button's minimum height is 60dp, not too sure if this is the same thing here?

components-compose/src/main/java/uk/gov/hmrc/components/compose/atom/button/Button.kt
```diff
@@ -15,4 +15,119 @@
...
+    textAlign: TextAlign = TextAlign.Center,
+    buttonColors: ButtonColors = ButtonDefaults.buttonColors(
+        containerColor = Color.Transparent, contentColor = colors.hmrcBlue
+    ),
+    onClick: () -> Unit,
+    content: @Composable RowScope.() -> Unit = {},
+) {
+    CompositionLocalProvider(LocalRippleTheme provides HmrcRippleTheme) {
+        HmrcButton(
+            text = text, modifier = modifier, contentPadding = PaddingValues(all = hmrc_spacing_16),
```
ngoulongkam, line 107: The padding for the current `SecondaryButton` is not al 16dp, can you check if this is appearing the same?

components-compose/src/main/java/uk/gov/hmrc/components/compose/atom/button/Button.kt
```diff
@@ -15,4 +15,119 @@
...
+            enabled = enabled, textAlign = textAlign, buttonColors = buttonColors, onClick = onClick,
+        ) {
+            content.invoke(this)
+        }
+    }
+}
+
+@Composable
+fun IconButton(
```
ngoulongkam, line 116: I cant see anything on `IconButton` that sets the icon colour, is it needed?

ngoulongkam (changes requested):

components-compose/src/main/java/uk/gov/hmrc/components/compose/atom/button/Button.kt
sajinabraham: to adjust with the buttons we used padding as 16dp, so SecondaryButton also shows the same behaviour as in xml one

components-compose/src/main/java/uk/gov/hmrc/components/compose/atom/button/Button.kt
sajinabraham: because icon button Inherited from Seconday button so not colour is required ?

ngoulongkam: does that also inherit the colour for the icon?

sajinabraham: yes think so , did you check the button on phone it shows the same behaviour as secondary button, so is it necessary to add colour to it? because just one icon is added to Icon button and it toggle with the dark and night mode as-well?

ngoulongkam: that's all good 👍 as long as it is inherited. 

I didnt download the app to look at it, I just compared it to the existing code.

ngoulongkam (approved):

PR #109, opened 2023-02-28
HMA-5067: Compose BulletedTextView

ngoulongkam:
# 📝 Description

Github Issue
  
- [ ] Updated CHANGELOG
- [ ] Updated README

# 🛠 Testing Notes

# 📸 Screenshots
HMA-5067 (1)
HMA-5067 (2)
HMA-5067 (3)
HMA-5067 (4)
HMA-5067 (5)

### DIFF
--- a/components-compose/src/main/java/uk/gov/hmrc/components/compose/atom/bullet/BulletedTextView.kt
+++ b/components-compose/src/main/java/uk/gov/hmrc/components/compose/atom/bullet/BulletedTextView.kt
@@ -1,0 +1,48 @@
+/*
+ * Copyright 2023 HM Revenue & Customs
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package uk.gov.hmrc.components.compose.atom.bullet
+
+import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.Row
+import androidx.compose.foundation.layout.padding
+import androidx.compose.material3.Text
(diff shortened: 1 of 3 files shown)


components-compose/src/main/java/uk/gov/hmrc/components/compose/ui/theme/HmrcTypography.kt
```diff
@@ -67,5 +67,11 @@ data class HmrcTypography(
         fontWeight = FontWeight.Normal,
         fontSize = 16.sp,
         color = hmrcRed
+    ),
+    val bulletedText: TextStyle = TextStyle(
+        fontFamily = FontFamily.Default,
+        fontWeight = FontWeight.Normal,
+        fontSize = 16.sp,
+        color = hmrcBlack
     )
```
jemmaSlater, line 76: Do we need to define this as it's own typography item? It's the same as the `body` text and we would always want it to be consistent with the body text I think? If we used the `bulletedText` as defined here in the code somewhere, it wouldn't come with a bullet so may be a little misleading? I think we can just use `body` text in the bullet component.

components-compose/src/main/java/uk/gov/hmrc/components/compose/atom/bullet/BulletedTextView.kt
```diff
@@ -0,0 +1,48 @@
...
+
+@Composable
+fun BulletedTextView(
+    text: String,
+    modifier: Modifier = Modifier,
+) {
+    Column {
+        Row(modifier = modifier) {
+            Text(
+                modifier = Modifier.padding(end = 16.dp),
```
jemmaSlater, line 36: Can we use the defined spacers?

```suggestion
                modifier = Modifier.padding(end = hmrc_spacing_16),
```

components-compose/src/main/java/uk/gov/hmrc/components/compose/atom/bullet/BulletedTextView.kt
```diff
@@ -0,0 +1,48 @@
...
+                text = "\u25CF",
+                style = HmrcTheme.typography.bulletedText,
+                textAlign = TextAlign.Start
+            )
+            Text(
+                text = text,
+                style = HmrcTheme.typography.bulletedText,
```
sajinabraham, line 43: i think style can be use like style = typography.bulletedText
instead of
style = HmrcTheme.typography.bulletedText

jemmaSlater (approved):

PR #120, opened 2023-04-24
HMA 5076 jetpack compose components multi column row view (molecule)

aliahmedbd:
# 📝 Description
HMA 5076 jetpack composes components with multi-column row view (molecule)

# 📸 Screenshots
Screenshot 2023-04-24 at 15 53 10


emmaatkins (approved):

sample-compose-fragments/src/main/java/uk/gov/hmrc/sample_compose_fragments/presentation/fragment/molecules/MultiColumnRowViewFragment.kt
```diff
@@ -0,0 +1,52 @@
...
+class MultiColumnRowViewFragment : Fragment() {
+
+    private val arg : MultiColumnRowViewFragmentArgs by navArgs()
+
+    override fun onCreateView(
+        inflater: LayoutInflater, container: ViewGroup?,
+        savedInstanceState: Bundle?
+    ): View {
+        // Inflate the layout for this fragment
+        return ComposeView(requireContext()).apply {
```
jemmaSlater, line 41: Although this is correct, the main goal for our "over-engineered sample" app here is to try and provide a good example of how we might integrate compose with our existing fragment/xml codebase to ease migration. So for the other components we've used a very simple xml file (`R.layout.fragment_compose_example`) which just contains a compose view in a linear layout to use as an example. We can access it with the binding like we currently do in our main app. I think it'd be nice to keep that consistent for now (although presumably when we do get to the point to start adding brand new screens in the main app with compose we may be able to do it this way, so there may be some value in having both quite soon.. for now I think consistency aids learning). Happy to hear thoughts if you disagree! :)

sample-compose-fragments/src/main/java/uk/gov/hmrc/sample_compose_fragments/presentation/fragment/molecules/MultiColumnRowViewFragment.kt
```diff
@@ -0,0 +1,52 @@
...
+            setContent {
```
jemmaSlater, line 42: I think we need to include `setViewCompositionStrategy(ViewCompositionStrategy.DisposeOnViewTreeLifecycleDestroyed)` here to prevent memory leaks, because we are using Fragments still. https://developer.android.com/jetpack/compose/migrate/interoperability-apis/compose-in-views#composition-strategy

sample-compose-fragments/src/main/java/uk/gov/hmrc/sample_compose_fragments/presentation/fragment/molecules/MultiColumnRowViewFragment.kt
```diff
@@ -0,0 +1,52 @@
...
+import dagger.hilt.android.AndroidEntryPoint
+import uk.gov.hmrc.components.compose.ui.theme.HmrcTheme
+import uk.gov.hmrc.components.compose.ui.theme.rememberWindowSizeClass
+import uk.gov.hmrc.sample_compose_fragments.presentation.screens.molecules.MultiColumnRowViewScreen
+import uk.gov.hmrc.sample_compose_fragments.presentation.screens.sampletemplate.HmrcSurface
+
+@AndroidEntryPoint
+class MultiColumnRowViewFragment : Fragment() {
+
+    private val arg : MultiColumnRowViewFragmentArgs by navArgs()
```
jemmaSlater, line 34: Do we need this? Doesn't look like it's used anywhere..

sample-compose-fragments/src/main/res/values/strings.xml
```diff
@@ -330,5 +330,7 @@
     <string name="menu_panel_example_3_body">A savings account with bonuses designed to help you start saving</string>
     <string name="menu_panel_example_4_title">News</string>
     <string name="menu_panel_example_4_body">News and current tax related information from HMRC</string>
+    <!-- TODO: Remove or change this placeholder text -->
+    <string name="hello_blank_fragment">Hello blank fragment</string>
```
jemmaSlater, line 334: Should this be removed?

sample-compose-fragments/src/main/res/navigation/nav_graph_molecules.xml
```diff
@@ -77,4 +81,16 @@
...
     </fragment>
 
+    <fragment
+        android:id="@+id/navMultiRowTextFragment"
+        android:label="@string/molecules_multi_column_row"
+        android:name="uk.gov.hmrc.sample_compose_fragments.presentation.fragment.molecules.MultiColumnRowViewFragment">
+
+        <argument
+            android:name="title"
+            app:argType="string"/>
```
jemmaSlater, line 91: I don't think we need these?

components-compose/src/main/java/uk/gov/hmrc/components/compose/molecule/multiColumnRowView/MultiColumnRowItem.kt
```diff
@@ -0,0 +1,25 @@
...
+ */
+package uk.gov.hmrc.components.compose.molecule.multiColumnRowView
+
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.text.TextStyle
+
+data class MultiColumnRowItem(
+    val text: String = "",
+    val modifier: Modifier = Modifier,
+    val textStyle: TextStyle
```
jemmaSlater, line 24: Usually we just want the text as standard body text. I think we can add that as a default here so we don't have to add it every time, just in cases where we want something different.
```suggestion
    val textStyle: TextStyle = HmrcTheme.typography.body
```

components-compose/src/main/java/uk/gov/hmrc/components/compose/molecule/multiColumnRowView/MultiColumnRowView.kt
```diff
@@ -0,0 +1,36 @@
...
+package uk.gov.hmrc.components.compose.molecule.multiColumnRowView
+
+import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.Row
+import androidx.compose.material3.Text
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Modifier
+
+@Composable
+fun MultiColumnRowView(
```
jemmaSlater, line 25: We know our MultiColumnRow should only have up to 3 columns, so can we restrict the usage a bit further to only allow at max 3 items? We also know the one closest to the right edge of the screen should always be right aligned, so we can build in some rules to the component to enforce that, so the consumer doesn't need to define it each time they use it. It would also be good if the consumer didn't need to define the weight of each column item each time, because the component should be able to determine that itself based on the number of columns.

See here for visual examples: 
Screenshot 2023-04-25 at 12 30 28

https://hmrc.invisionapp.com/console/share/EXOVJCPYDPN/332983439

sample-compose-fragments/src/main/java/uk/gov/hmrc/sample_compose_fragments/presentation/fragment/molecules/MultiColumnRowViewFragment.kt
aliahmedbd: Yeah, I got your point.

components-compose/src/main/java/uk/gov/hmrc/components/compose/molecule/multiColumnRowView/MultiColumnRowView.kt
```diff
@@ -0,0 +1,59 @@
...
+    private const val WEIGHT_FULL = 1f
+    private const val WEIGHT_HALF = 0.5f
+
+    @Composable
+    operator fun invoke(
+        modifier: Modifier = Modifier,
+        columnList: List<MultiColumnRowItem>
+    ) {
+        Box(modifier = modifier) {
+            Row {
```
jemmaSlater, line 37: Is there a benefit to nesting the Row inside a Box when the Row is the only child?

components-compose/src/main/java/uk/gov/hmrc/components/compose/molecule/multiColumnRowView/MultiColumnRowView.kt
```diff
@@ -0,0 +1,59 @@
...
+                val restrictedColumns = columnList.take(restrictedColumnCount)
+                restrictedColumns.forEachIndexed { index, column ->
+
+                    val textAlign =
+                        if (restrictedColumns.size >= 2 && index >= 1) TextAlign.End
+                        else TextAlign.Start
```
jemmaSlater, line 43: If I've read this correctly, I think it could be simplified to `if (index == 0) TextAlign.Start else TextAlign.End` ?

components-compose/src/main/java/uk/gov/hmrc/components/compose/molecule/multiColumnRowView/MultiColumnRowView.kt
aliahmedbd: Yeah, you are right, removed the needless parent Box composable.

components-compose/src/main/java/uk/gov/hmrc/components/compose/molecule/multiColumnRowView/MultiColumnRowView.kt
aliahmedbd: Yeah, you are right, simplified the conditions.

jemmaSlater (approved):

ngoulongkam (approved):

build.gradle
```diff
@@ -1,8 +1,12 @@
+import uk.gov.hmrc.components.buildsrc.Kotlin
+import uk.gov.hmrc.components.buildsrc.Gradle
+
 // Top-level build file where you can add configuration options common to all sub-projects/modules.
 buildscript {
 
     ext {
         compose_version = '1.3.3'
```
sajinabraham, line 8: i think we can update the `compose_version` to `compose_version = '1.4.1'`

ref: https://github.com/hmrc/android-components/commit/c92c7ed5433de3a96376af359cb14626ef474332

please check this branch and few more libraries are available to update if needed

sample-compose-fragments/src/main/java/uk/gov/hmrc/sample_compose_fragments/presentation/screens/colors/ColorsListScreen.kt
sajinabraham, line 1: `val listItems by viewModel.colorItems.collectAsState()` can be replace with `val listItems by viewModel.colorItems.collectAsStateWithLifecycle()`

emmaatkins (dismissed):

components-compose/src/main/java/uk/gov/hmrc/components/compose/ui/theme/Theme.kt
```diff
@@ -90,6 +122,18 @@ object HmrcTheme {
...
+
+    val dimensions: Dimensions
+        @Composable
+        get() = LocalAppDimens.current
+
+    val orientation: Orientation
+        @Composable
+        get() = LocalOrientationMode.current
+
+    val fonScale: Float
```
jemmaSlater, line 134:
```suggestion
    val fontScale: Float
```

aliahmedbd: My mistake, I have renamed the variable name.

jemmaSlater (approved):

components-compose/src/main/java/uk/gov/hmrc/components/compose/ui/extensions/ModifierExt.kt
```diff
@@ -0,0 +1,22 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package uk.gov.hmrc.components.compose.ui.extensions
+
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.semantics.semantics
+
+fun Modifier.enableTalkBacMergeAccessibility() =
```
jemmaSlater, line 21:
```suggestion
fun Modifier.enableTalkBackMergeAccessibility() =
```

aliahmedbd: Function renamed.

jemmaSlater (approved):

PR #129, opened 2023-06-23
Hma 5084 jetpack compose components warning view molecule

aliahmedbd:
# 📝 Description
https://jira.tools.tax.service.gov.uk/browse/HMA-5084

# 📸 Screenshots
Screenshot 2023-06-23 at 11 32 24
Screenshot 2023-06-23 at 11 31 01


emmaatkins (dismissed):

ngoulongkam (dismissed):

components-compose/src/main/res/drawable/components_select_row_circle.xml
```diff
@@ -0,0 +1,4 @@
+<selector xmlns:android="http://schemas.android.com/apk/res/android">
```
jemmaSlater, line 1: Do we need to add this?

components-compose/src/main/java/uk/gov/hmrc/components/compose/molecule/warningview/WarningView.kt
```diff
@@ -0,0 +1,94 @@
...
+                .fillMaxWidth()
+                .background(backgroundColor)
+                .enableTalkBackMergeAccessibility()
+        ) {
+            Row(
+                modifier = Modifier
+                    .fillMaxWidth()
+                    .padding(padding)
+
+            ) {
```
jemmaSlater, line 69: Why do we need to nest the Row inside a Box here? Is it needed for the accessibility?

sample-compose-fragments/src/main/java/uk/gov/hmrc/sample_compose_fragments/presentation/screens/molecules/WarningViewScreen.kt
```diff
@@ -0,0 +1,84 @@
...
+
+                HmrcCard(
+                    modifier = Modifier.padding(bottom = HmrcTheme.dimensions.hmrcSpacing16),
+                ) {
+                    WarningView(
+                        text = stringResource(id = R.string.longest_text),
+                        iconTintColor = HmrcTheme.colors.hmrcWhite,
+                        backgroundColor = HmrcTheme.colors.hmrcBlue,
+                        textColor = HmrcTheme.colors.hmrcWhite
+                    )
```
jemmaSlater, line 67: In this example, can we demonstrate how you can replace the icon? The xml sample used `app:icon="@drawable/ic_info"`

components-compose/src/main/res/drawable/components_select_row_circle.xml
aliahmedbd: No, actually these are not needed currently. I have copied them from the component module maybe we need these later.

components-compose/src/main/java/uk/gov/hmrc/components/compose/molecule/warningview/WarningView.kt
aliahmedbd: Yes, it's required because it makes padding, alignment and background easier.

components-compose/src/main/res/drawable/components_select_row_circle.xml
jemmaSlater: In general we should try and avoid adding stuff before we need it, as if for whatever reason we don't end up needing it then we've got unused files we need to go back and tidy up at some point. I think we can just add them when we need them and remove them for now. :)

jemmaSlater (approved):

PR #136, opened 2023-07-03
Hma 5079 android jetpack compose components status view

aliahmedbd:
# 📝 Description
https://jira.tools.tax.service.gov.uk/browse/HMA-5079

# 📸 Screenshots
screenshot1
screenshot2


components-compose/src/main/java/uk/gov/hmrc/components/compose/molecule/statusview/StatusView.kt
```diff
@@ -0,0 +1,141 @@
...
+            }
+            //endregion
+
+            //region Title
+            Text(
+                text = title,
+                style = HmrcTheme.typography.h5.copy(color = textColor),
+                modifier = Modifier.fillMaxSize(),
+                textAlign = TextAlign.Center
+            )
```
ngoulongkam, line 94: Would it work by using `Heading5` here?

components-compose/src/main/java/uk/gov/hmrc/components/compose/molecule/statusview/StatusView.kt
```diff
@@ -0,0 +1,141 @@
...
+                        },
+                    textAlign = bodyAlignment.alignment
+                )
+            }
+            //endregion
+
+            //region Primary Button
+            if (primaryButtonText.isNotEmpty()) {
+                Spacer(modifier = Modifier.height(HmrcTheme.dimensions.hmrcSpacing8))
+                PrimaryButton(text = "Primary Button") {
```
ngoulongkam, line 116: Hardcoded text.

sample-compose-fragments/src/main/res/drawable/ic_maintenance.xml
```diff
@@ -0,0 +1,30 @@
+<!--
+     Copyright 2018 HM Revenue & Customs
```
ngoulongkam, line 2: 2018.

ngoulongkam (changes requested):

components-compose/src/main/java/uk/gov/hmrc/components/compose/molecule/statusview/StatusView.kt
aliahmedbd: Yes we can, but one problem I am facing Heading5 has no style parameter to change text colour.

ngoulongkam (dismissed):

ngoulongkam (dismissed):

components-compose/src/main/java/uk/gov/hmrc/components/compose/molecule/statusview/StatusView.kt
jemmaSlater: One thing I think we lose here by not using the Heading component is the talkback announcement.
Can we use the base Heading component and provide the style directly instead? 

```
Heading(text = text, style = typography.h5.copy(color = textColor), modifier = modifier)
```

Is there another way we can achieve the text alignment while using this component still? If this isn't possible, we should just add the semantics here instead.
```
modifier = modifier.semantics { heading() }
```

sample-compose-fragments/src/main/java/uk/gov/hmrc/sample_compose_fragments/presentation/screens/molecules/StatusViewScreen.kt
```diff
@@ -0,0 +1,111 @@
...
+import uk.gov.hmrc.sample_compose_fragments.presentation.screens.sampletemplate.ExamplesSlot
+import uk.gov.hmrc.sample_compose_fragments.presentation.screens.sampletemplate.HmrcCard
+import uk.gov.hmrc.sample_compose_fragments.presentation.screens.sampletemplate.PlaceholderSlot
+import uk.gov.hmrc.sample_compose_fragments.presentation.screens.sampletemplate.ScreenScrollViewColumn
+
+object StatusViewScreen {
+
+    @Composable
+    operator fun invoke() {
+        ScreenScrollViewColumn {
```
jemmaSlater, line 34: Is there any benefit to setting the Screen up like this with the operator function, rather than just making a top level composable function?

sample-compose-fragments/src/main/res/drawable/ic_info.xml
```diff
@@ -1,6 +1,6 @@
 <vector xmlns:android="http://schemas.android.com/apk/res/android"
-        android:width="24dp"
-        android:height="24dp"
+        android:width="100dp"
+        android:height="100dp"
```
jemmaSlater, line 3: Just check this isn't going to affect the existing usage of this icon?

sample-compose-fragments/src/main/java/uk/gov/hmrc/sample_compose_fragments/presentation/screens/molecules/StatusViewScreen.kt
aliahmedbd: Nothing but It's more like an OOP structure, as these are reusable components so we can provide their own constant properties within the class(object). One advantage of this approach is that when we identify our own views, they will visually stand out from normal composable functions.

components-compose/src/main/java/uk/gov/hmrc/components/compose/molecule/statusview/StatusView.kt
```diff
@@ -56,7 +56,7 @@ object StatusView {
         bodyContentDesc: String = "",
         bodyAlignment: BodyAlignment = BodyAlignment.CENTER,
         textColor: Color = HmrcTheme.colors.hmrcBlack,
-        icon: Int,
+        icon: Int = R.drawable.ic_maintenance,
```
jemmaSlater, line 59: I don't think we need to provide a default value here.. it's just a required field for all consumers. The status view can be used for many scenarios and we don't really want to link it to the maintenance status at the component level.

components-compose/src/main/res/drawable/ic_maintenance.xml
```diff
@@ -0,0 +1,30 @@
+<!--
```
jemmaSlater, line 1: Please can we remove this file too now it's not being used? :)

jemmaSlater (approved):

ngoulongkam (approved):

PR #137, opened 2023-07-12
Hma 5077 android jetpack compose components select row view

aliahmedbd:
# 📝 Description
https://jira.tools.tax.service.gov.uk/browse/HMA-5077

# Test Notes
Talkback has been updated in alignment with the Material3 guideline. We haven't included any custom talkback; instead, we are using the system's default talkback.

# 📸 Screenshots
Screenshot 2023-07-12 at 15 28 39


components-compose/src/main/java/uk/gov/hmrc/components/compose/molecule/selectrow/SelectRowView.kt
```diff
@@ -0,0 +1,129 @@
...
+                    modifier = Modifier
+                        .selectable(
+                            selected = (selectedValue.value == index),
+                            onClick = {
+                                selectedValue.value = index
+                                onRowSelected(index, selectRow)
+                            },
+                            role = Role.RadioButton
+                        )
+                        .padding(HmrcTheme.dimensions.hmrcSpacing8)
```
jemmaSlater, line 91: Can we make sure the touch state ripple abides to material design principles and extends to each side of the card? At the moment, it get's cut off before the edge of the card.

sample-compose-fragments/src/main/java/uk/gov/hmrc/sample_compose_fragments/presentation/screens/molecules/SelectRowViewScreen.kt
```diff
@@ -0,0 +1,96 @@
...
+                            "Second row",
+                            "Show an error",
+                            stringResource(id = R.string.longer_text)
+                        ),
+                        errorText = errorText.value,
+                        initialSelection = 0
+                    ) { position, value ->
+                        if (position == THIRD_ROW_VIEW - 1) {
+                            errorText.value = "This is an error"
+                        }
```
jemmaSlater, line 67: This error state isn't being retained on device rotation. I think we will need to store it in a view model.

We can also improve the sample so that the error is only showing while the "show an error" option is selected. I think we should dismiss the error when one of the other options is selected. (this wasn't the case in the xml sample, but I think it makes more sense as an example!)

sample-compose-fragments/src/main/res/drawable/components_select_row_tick_unchecked.xml
```diff
@@ -0,0 +1,12 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
```
jemmaSlater, line 1: This file isn't being used

sample-compose-fragments/src/main/res/navigation/nav_graph_molecules.xml
```diff
@@ -115,6 +118,10 @@
         android:id="@+id/warningViewFragment"
         android:label="@string/molecules_warning"
         android:name="uk.gov.hmrc.sample_compose_fragments.presentation.fragment.molecules.WarningViewFragment" />
+    <fragment
+        android:id="@+id/selectRowViewFragment"
+        android:name="uk.gov.hmrc.sample_compose_fragments.presentation.fragment.molecules.SelectRowViewFragment"
+        android:label="@string/molecules_select_row" />
```
jemmaSlater, line 125:
```suggestion

    <fragment
        android:id="@+id/selectRowViewFragment"
        android:name="uk.gov.hmrc.sample_compose_fragments.presentation.fragment.molecules.SelectRowViewFragment"
        android:label="@string/molecules_select_row" />
```

components-compose/src/main/java/uk/gov/hmrc/components/compose/molecule/selectrow/SelectRowView.kt
```diff
@@ -0,0 +1,129 @@
...
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.res.painterResource
+import androidx.compose.ui.semantics.Role
+import androidx.compose.ui.unit.dp
+import uk.gov.hmrc.components.compose.R
+import uk.gov.hmrc.components.compose.ui.theme.HmrcTheme
+
+object SelectRowView {
+
+    private const val RESTRICTED_ITEM_COUNT = 4
```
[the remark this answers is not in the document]
aliahmedbd: Hi Jemma, In the XML version in `SelectRowGroup` I can see 4 `SelectRowView` added that's why I thought it may be restricted to only 4 items.

sample-compose-fragments/src/main/res/navigation/nav_graph_molecules.xml
aliahmedbd: Jemma, I didn't get this suggestion, seems the same to me.

components-compose/src/main/java/uk/gov/hmrc/components/compose/molecule/selectrow/SelectRowView.kt
```diff
@@ -0,0 +1,129 @@
...
+                    Text(
+                        text = selectRow,
+                        modifier = Modifier.fillMaxWidth(),
+                        style = HmrcTheme.typography.body
+                    )
+                }
+
+//                if (index < selectRowViewItems.size - 1) {
+//                    Spacer(modifier = Modifier.height(HmrcTheme.dimensions.hmrcSpacing16))
+//                }
```
ngoulongkam, line 125: do you need this commented code?

sample-compose-fragments/src/main/java/uk/gov/hmrc/sample_compose_fragments/presentation/screens/molecules/SelectRowViewScreen.kt
```diff
@@ -0,0 +1,108 @@
...
+                ) { position, value ->  // Handle the SelectRowView item click listener
+                    viewModel.setPlaceHolderRowSelectedPosition(position)
+                }
+            }
+            //endregion
+
+            ExamplesSlot {
+                //region Example one
+                HmrcCard(modifier = Modifier.padding(bottom = HmrcTheme.dimensions.hmrcSpacing16)) {
+                    //val errorText = remember { mutableStateOf("") }
```
ngoulongkam, line 63: Do you need this commented code?

ngoulongkam (changes requested):

ngoulongkam (approved):

components-compose/src/main/java/uk/gov/hmrc/components/compose/molecule/selectrow/SelectRowView.kt
```diff
@@ -0,0 +1,125 @@
...
+                            indication = rememberRipple(bounded = false),
+                            onClick = {}
+                        )
+                        .selectable(
+                            selected = (rowSelectedPosition == index),
+                            onClick = {
+                                onRowSelected(index, selectRow)
+                            },
+                            role = Role.RadioButton,
+                        )
```
jemmaSlater, line 91: Do we need both `clickable` and `selectable` here? I think `selectable` is an enhanced version of `clickable`.

components-compose/src/main/java/uk/gov/hmrc/components/compose/molecule/selectrow/SelectRowView.kt
```diff
@@ -0,0 +1,125 @@
...
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.res.painterResource
+import androidx.compose.ui.semantics.Role
+import uk.gov.hmrc.components.compose.R
+import uk.gov.hmrc.components.compose.ui.theme.HmrcTheme
+
+object SelectRowView {
+
+    const val NO_ROW_SELECTED_POSITION = -1
+    const val FIRST_ROW_SELECTED_POSITION = 0
```
jemmaSlater, line 45: This appears to only be used in the sample app. I don't think it needs to be included in the component itself?

sample-compose-fragments/src/main/java/uk/gov/hmrc/sample_compose_fragments/presentation/screens/molecules/SelectRowViewScreen.kt
```diff
@@ -0,0 +1,107 @@
...
+                            stringResource(R.string.select_row_view_first_row),
+                            stringResource(R.string.select_row_view_second_row),
+                            stringResource(R.string.select_row_view_error_row),
+                            stringResource(id = R.string.longer_text)
+                        ),
+                        errorText = exampleUiState.errorTextExample1,
+                        rowSelectedPosition = exampleUiState.rowSelectedPositionExample1
+                    ) { position, value -> // Handle the SelectRowView item click listener
+                        viewModel.setRowSelectedPositionExample1(position)
+                        if (position == THIRD_ROW_VIEW - 1) {
```
jemmaSlater, line 74: This is the only usage of this const, so we could just set the const to the index of the item we want so we don't need to do additional calculations here.

```suggestion
           private const val ERROR_ITEM_INDEX = 2
                        if (position == ERROR_ITEM_INDEX) {
```

sample-compose-fragments/src/main/java/uk/gov/hmrc/sample_compose_fragments/util/ComposeUtil.kt
```diff
@@ -0,0 +1,59 @@
...
+import androidx.compose.ui.platform.LocalLifecycleOwner
+import androidx.lifecycle.Lifecycle
+import androidx.lifecycle.LifecycleOwner
+import androidx.lifecycle.flowWithLifecycle
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.StateFlow
+import kotlin.coroutines.CoroutineContext
+import kotlin.coroutines.EmptyCoroutineContext
+
+object ComposeUtil {
```
jemmaSlater, line 31: Do we need these custom extension functions? Can we use the `.collectAsStateWithLifecycle()` function instead? https://developer.android.com/jetpack/compose/state#use-other-types-of-state-in-jetpack-compose

sample-compose-fragments/src/main/res/navigation/nav_graph_molecules.xml
jemmaSlater: Sorry, I was just being picky over the spacing, no code changes needed!

components-compose/src/main/java/uk/gov/hmrc/components/compose/molecule/selectrow/SelectRowView.kt
aliahmedbd: For anyone using that component, I thought this const would be more clear.

sample-compose-fragments/src/main/java/uk/gov/hmrc/sample_compose_fragments/util/ComposeUtil.kt
aliahmedbd: Yeah we can use `.collectAsStateWithLifecycle()` that function, one minor benefit I can see:

The benefit of using `rememberFlow` and `collectAsStateLifecycleAware` instead of `collectAsStateWithLifecycle` is that it allows you to reuse the flow across multiple composables. This is because `rememberFlow` creates a new flow that is scoped to the lifecycle of the composable that it is used in. This means that the flow will only be active while the composable is visible, and it will be automatically disposed of when the composable is destroyed.

`collectAsStateWithLifecycle` on the other hand, creates a new flow for each composable that it is used in. This means that if you use `collectAsStateWithLifecycle` in multiple composables, you will end up with multiple flows that are all tracking the same data. This can lead to memory leaks and other problems.

components-compose/src/main/java/uk/gov/hmrc/components/compose/molecule/selectrow/SelectRowView.kt
```diff
@@ -0,0 +1,120 @@
...
+     * @param errorText Error text to be displayed, if any.
+     * @param onRowSelected Callback function triggered when an item is clicked.
+     */
+    @Composable
+    operator fun invoke(
+        modifier: Modifier = Modifier,
+        selectRowViewItems: List<String>,
+        checkedIcon: Int = R.drawable.components_select_row_circle_checked,
+        uncheckedIcon: Int = R.drawable.components_select_row_circle_unchecked,
+        rowSelectedPosition: Int,
```
jemmaSlater, line 60: I think `rowSelectedPosition` either needs to be a nullable parameter, or we should reintroduce the default parameter with the `const val NO_ROW_SELECTED_POSITION = -1`. 

My previous comment about these consts was specifically about not needing the `FIRST_ROW_SELECTED_POSITION` const for the first value, as it didn't really make sense to define that position explicitly but none of the others.. I don't think we need to provide consts for the items, but a default state for nothing selected seems useful. There will be common cases where we don't have anything pre-selected for the user.

jemmaSlater (dismissed):

jemmaSlater (dismissed):

components-compose/src/main/res/values/strings.xml
```diff
@@ -18,4 +18,7 @@
     <string name="textInputView_error">Error</string>
     <string name="textInputView_clear">Clear text</string>
     <string name="accessibility_warning">warning; %s</string>
+    <string name="accessibility_selected">selected</string>
+    <string name="accessibility_not_selected">not selected</string>
+    <string name="accessibility_radio_button">radio button</string>
```
aliahmedbd, line 23: Can we remove these strings now?

ngoulongkam (approved):

jemmaSlater (dismissed):

PR #146, opened 2023-09-19
HMA-5085: Compose for IconButtonCardView component

ngoulongkam:
# 📝 Description

Github Issue
https://jira.tools.tax.service.gov.uk/browse/HMA-5085
  
- [ ] Updated CHANGELOG
- [ ] Updated README

# 🛠 Testing Notes

# 📸 Screenshots
HMA-5085

### DIFF
--- a/components-compose/src/main/java/uk/gov/hmrc/components/compose/organism/ComposableTemplate.kt
+++ b/components-compose/src/main/java/uk/gov/hmrc/components/compose/organism/ComposableTemplate.kt
@@ -27,12 +27,12 @@ import uk.gov.hmrc.components.compose.ui.theme.HmrcTheme
 
 @Composable
 fun HmrcCardView(
-    modifier: Modifier? = Modifier,
+    modifier: Modifier = Modifier,
     elevation: CardElevation = CardDefaults.cardElevation(),
     content: @Composable ColumnScope.() -> Unit
 ) {
     Card(
-        modifier = modifier?.fillMaxWidth() ?: Modifier.fillMaxWidth(),
+        modifier = modifier.fillMaxWidth(),
         shape = RectangleShape,
         content = content,
         colors = CardDefaults.cardColors(containerColor = HmrcTheme.colors.hmrcWhiteBackground),
--- a/components-compose/src/main/java/uk/gov/hmrc/components/compose/organism/ComposableTemplate.kt
+++ b/components-compose/src/main/java/uk/gov/hmrc/components/compose/organism/ComposableTemplate.kt
@@ -1,0 +1,37 @@
+/*
+ * Copyright 2023 HM Revenue & Customs
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package uk.gov.hmrc.components.compose.organism
+
+import androidx.compose.foundation.layout.ColumnScope
+import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.material3.Card
+import androidx.compose.material3.CardDefaults
+import androidx.compose.material3.CardElevation
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.graphics.RectangleShape
+import uk.gov.hmrc.components.compose.ui.theme.HmrcTheme
+
+@Composable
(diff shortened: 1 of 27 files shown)


components-compose/src/main/java/uk/gov/hmrc/components/compose/organism/ComposableTemplate.kt
```diff
@@ -0,0 +1,41 @@
...
+import uk.gov.hmrc.components.compose.ui.theme.HmrcTheme
+
+@Composable
+fun HmrcCardView(
+    modifier: Modifier? = Modifier,
+    elevation: CardElevation = CardDefaults.cardElevation(),
+    content: @Composable ColumnScope.() -> Unit
+) {
+    Card(
+        modifier = modifier?.fillMaxWidth() ?: Modifier.fillMaxWidth(),
```
jemmaSlater, line 35: Does `modifier` need to be a nullable parameter here? We have a non-null default parameter, and if null was provided, all we do is default to Modifier anyway which is the same as the default.. 
```suggestion
fun HmrcCardView(
    modifier: Modifier = Modifier,
    elevation: CardElevation = CardDefaults.cardElevation(),
    content: @Composable ColumnScope.() -> Unit
) {
    Card(
        modifier = modifier.fillMaxWidth(),
```

sample-compose-fragments/src/main/java/uk/gov/hmrc/sample_compose_fragments/data/repository/RepositoryImp.kt
```diff
@@ -63,5 +69,8 @@ class RepositoryImpl @Inject constructor() : Repository {
         const val MOLECULE_STATUS_VIEW = 10
         const val MOLECULE_WARNING_VIEW = 11
         const val SELECT_ROW_VIEW = 12
+
+        // ORGANISM
+        const val ORGANISM_ICON_BUTTON_CARD_VIEW = 1
```
jemmaSlater, line 74: It feels a bit strange at the moment, but can we start with the intentions of having the compose sample app a replica of the existing xml sample app.. so the `IconButtonCardView` is 4th in the list... 
```suggestion
        const val ORGANISM_ICON_BUTTON_CARD_VIEW = 4
```

components-compose/src/main/java/uk/gov/hmrc/components/compose/organism/ComposableTemplate.kt
```diff
@@ -0,0 +1,41 @@
...
+import androidx.compose.material3.CardElevation
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.graphics.RectangleShape
+import uk.gov.hmrc.components.compose.ui.theme.HmrcTheme
+
+@Composable
+fun HmrcCardView(
+    modifier: Modifier = Modifier,
+    elevation: CardElevation = CardDefaults.cardElevation(),
```
jemmaSlater, line 31: Thinking about this even further, do we want the elevation to be customisable? Thinking we probably want to maintain a consistent elevation across all our `HmrcCard`s so leaning more towards this not being a parameter at all and just set in the card directly..?

components-compose/src/main/java/uk/gov/hmrc/components/compose/organism/ComposableTemplate.kt
```diff
@@ -0,0 +1,39 @@
...
+import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.material3.Card
+import androidx.compose.material3.CardDefaults
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.graphics.RectangleShape
+import uk.gov.hmrc.components.compose.ui.theme.HmrcTheme
+
+@Composable
+fun HmrcCardView(
```
aliahmedbd, line 28: There is another card view in `uk.gov.hmrc.sample_compose_fragments.presentation.screens.sampletemplate`
can we reuse it?

components-compose/src/main/java/uk/gov/hmrc/components/compose/organism/card/IconButtonCardView.kt
```diff
@@ -0,0 +1,38 @@
...
+import androidx.annotation.DrawableRes
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Modifier
+import uk.gov.hmrc.components.compose.atom.button.IconButton
+import uk.gov.hmrc.components.compose.organism.HmrcCardView
+
+@Composable
+fun IconButtonCardView(
+    titleText: String,
+    @DrawableRes iconResId: Int,
```
aliahmedbd, line 27: As the same icon resource we are using, can we make default iconResId ->  R.drawable.ic_help_outline,
@DrawableRes iconResId: Int = R.drawable.ic_help_outline

aliahmedbd: I agree with that -> _"for the default icon, Im going to push back on that, because its only the example we’re using ic_help_outline , but it should be up to the consumer app to set the icon, rather then using a default one."_

emmaatkins (approved):

aliahmedbd (approved):

PR #150, opened 2023-11-24
Hma 5086 android jetpack compose components   headline card view organism

aliahmedbd:
# 📝 Description
https://jira.tools.tax.service.gov.uk/browse/HMA-5086

# 🛠 Testing Notes
Organism -> Headline Card View

# 📸 Screenshots
Screenshot 2023-11-24 at 18 00 49


components-compose/src/main/java/uk/gov/hmrc/components/compose/organism/headline/HeadlineCardView.kt
```diff
@@ -0,0 +1,141 @@
...
+    operator fun invoke(
+        modifier: Modifier = Modifier,
+        headline: String,
+        title: String,
+        isClickable: Boolean = false,
+        childPadding: Boolean = true,
+        views: List<@Composable () -> Unit> = listOf(),
+        onHeadlineCardClick: () -> Unit = {}
+    ) {
+        if (isClickable) {
```
ngoulongkam, line 54: instead of providing both `isClickable` AND `onHeadlineCardClick`.

can we just use `onHeadlineCardClick` only, as I assume we'll never make this component clickable without `onHeadlineCardClick`. (this also lines up with how the component work previously).

components-compose/src/main/java/uk/gov/hmrc/components/compose/organism/headline/HeadlineCardView.kt
```diff
@@ -0,0 +1,141 @@
...
+                        .background(HmrcTheme.colors.hmrcWhiteBackground)
+                        .clickable {
+                            onHeadlineCardClick()
+                        }
+                ) {
+                    HeadlineCard(modifier = modifier.then(Modifier.weight(1f)), headline, title, childPadding, views)
+                    Image(
+                        modifier = Modifier.align(Alignment.CenterVertically),
+                        painter = painterResource(id = R.drawable.components_ic_chevron_right),
+                        contentDescription = ""
```
ngoulongkam, line 67: Im not 100% sure if this is where you need to put this.

Do we need to set the content description to `accessibility_button_activate` here like the old component?

ngoulongkam (changes requested):

ngoulongkam (dismissed):

components-compose/src/main/java/uk/gov/hmrc/components/compose/organism/headline/HeadlineCardView.kt
```diff
@@ -0,0 +1,157 @@
...
+    @Composable
+    operator fun invoke(
+        modifier: Modifier = Modifier,
+        headline: String,
+        title: String,
+        childPadding: Boolean = true,
+        views: List<@Composable () -> Unit> = listOf(),
+        chevronContentDescription: String = stringResource(id = R.string.accessibility_button_activate),
+        onHeadlineCardClick: (() -> Unit)? = null
+    ) {
```
jemmaSlater, line 55: Are we able to rearrange this so that the views are the last parameter and then we can use it as more of an adaptable Composable container, like we can add rows inside a Row or Column? Similar to the way we build up the IconButton component in the existing Compose components. I think this will be a better pattern for our "extendable" card organisms like PrimaryCard etc. and we can be consistent in usage across the board.

```suggestion
    operator fun invoke(
        modifier: Modifier = Modifier,
        headline: String,
        title: String,
        childPadding: Boolean = true,
        chevronContentDescription: String = stringResource(id = R.string.accessibility_button_activate),
        onHeadlineCardClick: (() -> Unit)? = null,
        content: @Composable ColumnScope.() -> Unit = {},
    ) {
```

Then usage can be more intuative with how it's visually displayed, eg.
```
HeadlineCardView(headline: "Headline", title: "Title") {
    Text("Extra text")
    InsetText("Helpful information")
    Button("Call to action")
}

File: components-compose/src/main/java/uk/gov/hmrc/components/compose/organism/headline/HeadlineCardView.kt [hunk 4]
```diff
@@ -0,0 +1,157 @@
...
+        }
+    }
+
+    @Composable
+    fun HeadlineCard(
+        modifier: Modifier = Modifier,
+        headline: String,
+        title: String,
+        childPadding: Boolean = true,
+        views: List<@Composable () -> Unit> = listOf()
```
jemmaSlater, line 97: I think we can restrict this to a ColumnScope as per my above suggestion. I think we previously added to a LinearLayout so we only need to support the linear vertical option! If we name it `content` then it's a bit more consistent with what we've already built too :) 

```suggestion
        content: @Composable ColumnScope.() -> Unit = {},
```

emmaatkins (dismissed):

components-compose/src/main/java/uk/gov/hmrc/components/compose/organism/headline/HeadlineCardView.kt
```diff
@@ -0,0 +1,212 @@
...
+                        .fillMaxWidth()
+                        .semantics(mergeDescendants = true) {
+                            isTraversalGroup = true
+                            traversalIndex = TRAVERSE_INDEX_3RD
+                        }
+                        .then(
+                            if (childPaddingState) {
+                                Modifier.padding(horizontal = HmrcTheme.dimensions.hmrcSpacing16)
+                            } else {
+                                Modifier.padding(0.dp)
```
jemmaSlater, line 166: Is 0 padding default anyway, does it need to be defined?

components-compose/src/main/java/uk/gov/hmrc/components/compose/organism/headline/HeadlineCardView.kt
```diff
@@ -0,0 +1,212 @@
...
+                                .semantics(mergeDescendants = true) {}
+                        ),
+                        headline, headlineContentDescription, title, titleContentDescription, childPadding, content
+                    )
+                    Image(
+                        modifier = Modifier
+                            .align(Alignment.CenterVertically)
+                            .semantics(mergeDescendants = true) {
+                                contentDescription = chevronContentDescription
+                                traversalIndex = TRAVERSE_INDEX_4TH
```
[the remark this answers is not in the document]
aliahmedbd: Jemma, appreciate your feedback. Upon investigation, it functions seamlessly when there are multiple contents. The entire content is designated as isTraversalGroup = true with a traverse index of traversalIndex = TRAVERSE_INDEX_3RD. This setup allows the content to work as a cohesive group, and once the full traverse is complete, it seamlessly transitions to the next traversal index.
I have attached the screen record for that.

https://github.com/hmrc/android-components/assets/11981999/2037345e-35ed-4686-b0b6-b2d79c6007c8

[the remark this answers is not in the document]
jemmaSlater: Great - thanks for checking! :)

jemmaSlater (approved):

ngoulongkam (approved):

PR #154, opened 2024-01-03
HMA-5089 PrimaryCardView Compose

emmaatkins:
# 📝 Description

Github Issue
https://jira.tools.tax.service.gov.uk/browse/HMA-5089
  
- [ ] Updated CHANGELOG
- [ ] Updated README

# 🛠 Testing Notes

# 📸 Screenshots

### DIFF
--- a/components-compose/src/main/java/uk/gov/hmrc/components/compose/organism/primary/PrimaryCardView.kt
+++ b/components-compose/src/main/java/uk/gov/hmrc/components/compose/organism/primary/PrimaryCardView.kt
@@ -15,4 +15,64 @@
  */
 package uk.gov.hmrc.components.compose.organism.primary
 
-class PrimaryCardView
+import androidx.compose.foundation.background
+import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.ColumnScope
+import androidx.compose.foundation.layout.Spacer
+import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.foundation.layout.height
+import androidx.compose.foundation.layout.padding
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.derivedStateOf
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.remember
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.semantics.contentDescription
+import androidx.compose.ui.semantics.semantics
+import uk.gov.hmrc.components.compose.atom.heading.Heading5
+import uk.gov.hmrc.components.compose.ui.theme.HmrcTheme
+
+@Composable
+fun PrimaryCard(
+    modifier: Modifier = Modifier,
+    title: String,
+    titleContentDescription: String = "",
+    childPadding: Boolean = true,
+    content: @Composable (ColumnScope.() -> Unit)? = null
+) {
+    val childPaddingState by remember { derivedStateOf { childPadding } }
+
+    Column(
+        modifier = modifier
+            .background(HmrcTheme.colors.hmrcWhiteBackground)
+            .fillMaxWidth()
+    ) {
+        Heading5(
+            text = title,
+            modifier = Modifier
+                .padding(horizontal = HmrcTheme.dimensions.hmrcSpacing16)
+                .padding(top = HmrcTheme.dimensions.hmrcSpacing16)
+                .semantics(mergeDescendants = true) {
+                    contentDescription = titleContentDescription
+                }
+                .fillMaxWidth()
+        )
+        if (content != null) {
+            Column(
+                modifier = Modifier
+                    .fillMaxWidth()
+                    .then(
+                        if (childPaddingState) {
+                            Modifier.padding(top = HmrcTheme.dimensions.hmrcSpacing16)
+                                .padding(horizontal = HmrcTheme.dimensions.hmrcSpacing16)
+                        } else {
+                            Modifier
+                        }
+                    )
+            ) {
+                content()
+            }
+            if (childPaddingState) Spacer(modifier = Modifier.height(HmrcTheme.dimensions.hmrcSpacing16))
+        }
+    }
+}
--- a/sample-compose-fragments/src/main/java/uk/gov/hmrc/sample_compose_fragments/presentation/screens/organisms/PrimaryCardViewScreen.kt
+++ b/sample-compose-fragments/src/main/java/uk/gov/hmrc/sample_compose_fragments/presentation/screens/organisms/PrimaryCardViewScreen.kt
@@ -1,0 +1,124 @@
+/*
+ * Copyright 2023 HM Revenue & Customs
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package uk.gov.hmrc.sample_compose_fragments.presentation.screens.organisms
+
+import android.widget.Toast
+import androidx.compose.foundation.layout.Spacer
+import androidx.compose.foundation.layout.height
+import androidx.compose.foundation.layout.padding
+import androidx.compose.material3.Text
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.res.stringResource
+import androidx.compose.ui.text.style.TextAlign
+import uk.gov.hmrc.components.compose.atom.button.IconButton
+import uk.gov.hmrc.components.compose.atom.button.PrimaryButton
+import uk.gov.hmrc.components.compose.atom.button.SecondaryButton
+import uk.gov.hmrc.components.compose.molecule.inset.InsetTextView
+import uk.gov.hmrc.components.compose.organism.HmrcCardView
+import uk.gov.hmrc.components.compose.organism.primary.PrimaryCard
+import uk.gov.hmrc.components.compose.ui.theme.HmrcTheme
+import uk.gov.hmrc.sample_compose_components.R
+import uk.gov.hmrc.sample_compose_fragments.presentation.screens.sampletemplate.ExamplesSlot
+import uk.gov.hmrc.sample_compose_fragments.presentation.screens.sampletemplate.PlaceholderSlot
+import uk.gov.hmrc.sample_compose_fragments.presentation.screens.sampletemplate.ScreenScrollViewColumn
+
+@Composable
+fun PrimaryCardViewScreen(onClickAction: () -> Unit) {
+    ScreenScrollViewColumn {
+        PlaceholderSlot {
+            PrimaryCard(
+                title = stringResource(id = R.string.primary_card_placeholder_title),
+                content = {
+                    Text(
+                        text = stringResource(id = R.string.primary_card_placeholder_body),
+                        style = HmrcTheme.typography.body
+                    )
+                }
+            )
+        }
+
+        ExamplesSlot {
+            HmrcCardView(modifier = Modifier.padding(bottom = HmrcTheme.dimensions.hmrcSpacing16)) {
+                PrimaryCard(
+                    title = stringResource(id = R.string.primary_card_example_1_title),
+                    content = {
+                        Text(
+                            text = stringResource(id = R.string.primary_card_example_1_body),
+                            style = HmrcTheme.typography.body
+                        )
+                    }
+                )
+            }
+            HmrcCardView(modifier = Modifier.padding(bottom = HmrcTheme.dimensions.hmrcSpacing16)) {
+                PrimaryCard(
+                    title = stringResource(id = R.string.primary_card_example_2_title),
+                    content = {
+                        Text(
+                            text = stringResource(id = R.string.primary_card_example_2_body),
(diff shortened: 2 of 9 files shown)


components-compose/src/main/java/uk/gov/hmrc/components/compose/organism/primary/PrimaryCardView.kt
```diff
@@ -15,4 +15,64 @@
...
+            .background(HmrcTheme.colors.hmrcWhiteBackground)
+            .fillMaxWidth()
+    ) {
+        Heading5(
+            text = title,
+            modifier = Modifier
+                .padding(horizontal = HmrcTheme.dimensions.hmrcSpacing16)
+                .padding(top = HmrcTheme.dimensions.hmrcSpacing16)
+                .semantics(mergeDescendants = true) {
+                    contentDescription = titleContentDescription
```
ngoulongkam, line 56: I may be wrong here, but if we set content description to empty string here as default, wouldn't this make talkback read out nothing? 

Do we need to check string is not empty, before setting the content description?

components-compose/src/main/java/uk/gov/hmrc/components/compose/organism/primary/PrimaryCardView.kt
```diff
@@ -15,4 +15,64 @@
...
+                }
+                .fillMaxWidth()
+        )
+        if (content != null) {
+            Column(
+                modifier = Modifier
+                    .fillMaxWidth()
+                    .then(
+                        if (childPaddingState) {
```
ngoulongkam, line 65: Should the `childPaddingState` set at a higher level, covering the whole component, rather then only covering the `content` part of the component?

components-compose/src/main/java/uk/gov/hmrc/components/compose/organism/primary/PrimaryCardView.kt
```diff
@@ -15,4 +15,64 @@
...
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.remember
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.semantics.contentDescription
+import androidx.compose.ui.semantics.semantics
+import uk.gov.hmrc.components.compose.atom.heading.Heading5
+import uk.gov.hmrc.components.compose.ui.theme.HmrcTheme
+
+@Composable
+fun PrimaryCard(
```
ngoulongkam, line 36: Can we match the old component `PrimaryCardView` name please

sample-compose-fragments/src/main/res/navigation/nav_graph.xml
```diff
@@ -59,5 +62,9 @@
         android:name="uk.gov.hmrc.sample_compose_fragments.presentation.fragment.organisms.HeadlineCardFragment"
         android:label="@string/organisms_headline_card"
         tools:layout="@layout/fragment_compose_example"/>
-
+    <fragment
+        android:id="@+id/primaryCardFragment"
```
ngoulongkam, line 66: Same here `primaryCardViewFragment`, matching the name of the fragment.

sample-compose-fragments/src/main/res/navigation/nav_graph.xml
```diff
@@ -33,6 +33,9 @@
         <action
             android:id="@+id/action_organismsFragment_to_headlineCardFragment"
             app:destination="@id/headlineCardFragment" />
+        <action
+            android:id="@+id/action_organismsFragment_to_primaryCardFragment"
+            app:destination="@id/primaryCardFragment" />
```
ngoulongkam, line 38: Same here `primaryCardViewFragment`, matching the name of the fragment.

sample-compose-fragments/src/main/res/values/strings.xml
```diff
@@ -368,4 +368,18 @@
     <string name="tab_bar_view_long_tab_item_3">and component becomes</string>
     <string name="tab_bar_view_long_tab_item_4">scrollable</string>
 
+    <!-- Primary Card -->
```
ngoulongkam, line 371: Primary Card View

ngoulongkam (changes requested):

components-compose/src/main/java/uk/gov/hmrc/components/compose/organism/primary/PrimaryCardView.kt
```diff
@@ -15,4 +15,64 @@
...
+    Column(
+        modifier = modifier
+            .background(HmrcTheme.colors.hmrcWhiteBackground)
+            .fillMaxWidth()
+    ) {
+        Heading5(
+            text = title,
+            modifier = Modifier
+                .padding(horizontal = HmrcTheme.dimensions.hmrcSpacing16)
+                .padding(top = HmrcTheme.dimensions.hmrcSpacing16)
```
jemmaSlater, line 54: Feels a bit odd to have two different padding modifiers declared.. can we combine them?

components-compose/src/main/java/uk/gov/hmrc/components/compose/organism/primary/PrimaryCardView.kt
```diff
@@ -15,4 +15,64 @@
...
+                .fillMaxWidth()
+        )
+        if (content != null) {
+            Column(
+                modifier = Modifier
+                    .fillMaxWidth()
+                    .then(
+                        if (childPaddingState) {
+                            Modifier.padding(top = HmrcTheme.dimensions.hmrcSpacing16)
+                                .padding(horizontal = HmrcTheme.dimensions.hmrcSpacing16)
```
jemmaSlater, line 67: Can we combine these paddings into one padding?

components-compose/src/main/java/uk/gov/hmrc/components/compose/organism/primary/PrimaryCardView.kt
```diff
@@ -15,4 +15,64 @@
...
+                        } else {
+                            Modifier
+                        }
+                    )
+            ) {
+                content()
+            }
+            if (childPaddingState) Spacer(modifier = Modifier.height(HmrcTheme.dimensions.hmrcSpacing16))
```
jemmaSlater, line 75: Can we not add this to the Column modifier?

sample-compose-fragments/src/main/java/uk/gov/hmrc/sample_compose_fragments/presentation/screens/organisms/PrimaryCardViewScreen.kt
```diff
@@ -0,0 +1,124 @@
...
+                        text = stringResource(id = R.string.primary_card_placeholder_body),
+                        style = HmrcTheme.typography.body
+                    )
+                }
+            )
+        }
+
+        ExamplesSlot {
+            HmrcCardView(modifier = Modifier.padding(bottom = HmrcTheme.dimensions.hmrcSpacing16)) {
+                PrimaryCard(
```
jemmaSlater, line 56: This card view should be built into the component.. the component should be a card view.. then we can remove this as we don't need to put a card on a card.

sample-compose-fragments/src/main/java/uk/gov/hmrc/sample_compose_fragments/presentation/screens/organisms/PrimaryCardViewScreen.kt
```diff
@@ -0,0 +1,124 @@
...
+            }
+            HmrcCardView(modifier = Modifier.padding(bottom = HmrcTheme.dimensions.hmrcSpacing16)) {
+                PrimaryCard(
+                    childPadding = false,
+                    title = stringResource(id = R.string.primary_card_example_2_title),
+                    content = {
+                        Text(
+                            modifier = Modifier
+                                .padding(horizontal = HmrcTheme.dimensions.hmrcSpacing16)
+                                .padding(top = HmrcTheme.dimensions.hmrcSpacing16),
```
jemmaSlater, line 101: Can we combine these paddings?

sample-compose-fragments/src/main/res/values/strings.xml
```diff
@@ -368,4 +368,18 @@
...
+    <string name="primary_card_placeholder_title">Title</string>
+    <string name="primary_card_placeholder_body">Body</string>
+    <string name="primary_card_example_1_title">Renew your tax credits</string>
+    <string name="primary_card_example_1_body">Renew your tax credits or check the progress of your claims</string>
+    <string name="primary_card_example_2_title">Lorem ipsum dolor sit amet, ne nec regione nostrum. Quo altera nusquam apeirian et, melius discere pro ne, sint unum.</string>
+    <string name="primary_card_example_2_body">Lorem ipsum dolor sit amet, his ne quis phaedrum. Illum eloquentiam in pro, aperiam conceptam complectitur cum in, eum paulo
+		ceteros at. Vel erroribus adversarium at, integre omnesque eu est. Appareat expetenda cum no. Erant omnes interesset pro at, id mei munere moderatius,
+		ius ne ignota perpetua.
+	</string>
```
jemmaSlater, line 380: Do we need to add these? Can we reuse `longest_text` and `longer_text`?

sample-compose-fragments/src/main/res/values/strings.xml
```diff
@@ -368,4 +368,18 @@
...
+    <string name="primary_card_example_3_button">Renew or check my claims</string>
+    <string name="primary_d_carexample_4_button">Lorem ipsum dolor sit amet, id cum ullum deseruisse solet.</string>
```
jemmaSlater, line 382: Do we need this? Can we just use `long_text` instead?

components-compose/src/main/java/uk/gov/hmrc/components/compose/organism/primary/PrimaryCardView.kt
jemmaSlater: I don't think so.. the child padding is meant for automatically adding padding to eg. text or buttons added in the content section of this card, but we need to be able to remove the default padding if we want to add eg. a secondary button that needs to stretch to the edges. 
So I think it's correct for the child padding to only apply to the children (content).

components-compose/src/main/java/uk/gov/hmrc/components/compose/organism/primary/PrimaryCardView.kt
emmaatkins: It goes after the content?

jemmaSlater: Isn't the Column the container for the content though? So I would have assumed it would work in a similar way to xml where the bottom padding of the layout would apply around the children?

components-compose/src/main/java/uk/gov/hmrc/components/compose/organism/primary/PrimaryCardView.kt
aliahmedbd: Kam, I have checked with and without a custom content description, and it seems to work perfectly.

aliahmedbd (approved):

components-compose/src/main/java/uk/gov/hmrc/components/compose/organism/headline/HeadlineCardView.kt
```diff
@@ -160,15 +160,14 @@ object HeadlineCardView {
                         }
                         .then(
                             if (childPaddingState) {
-                                Modifier.padding(horizontal = HmrcTheme.dimensions.hmrcSpacing16)
+                                Modifier.padding(HmrcTheme.dimensions.hmrcSpacing16)
```
jemmaSlater, line 163: Do you need to remove the Spacer above now that you're adding top padding here too?

components-compose/src/main/java/uk/gov/hmrc/components/compose/organism/primary/PrimaryCardView.kt
```diff
@@ -15,4 +15,60 @@
...
+                        } else {
+                            Modifier
+                        }
+                    )
+            ) {
+                content()
+            }
+        }
+    }
+}
```
jemmaSlater, line 74: Please can you add a couple of Preview functions to show the states with the minimum and the maximum? eg. one without any content and one with some content?

sample-compose-fragments/src/main/java/uk/gov/hmrc/sample_compose_fragments/presentation/screens/organisms/PrimaryCardViewScreen.kt
```diff
@@ -0,0 +1,122 @@
...
+                modifier = Modifier.padding(bottom = HmrcTheme.dimensions.hmrcSpacing16),
+                title = stringResource(id = R.string.primary_card_example_1_title),
+                content = {
+                    Text(
+                        text = stringResource(id = R.string.primary_card_example_1_body),
+                        style = HmrcTheme.typography.body
+                    )
+                }
+            )
+        }
```
jemmaSlater, line 63: I think we want all of the examples in the ExamplesSlot block, not just the first one

components-compose/src/main/java/uk/gov/hmrc/components/compose/organism/primary/PrimaryCardView.kt
```diff
@@ -15,4 +15,96 @@
...
+            title = "",
+            content = {
+                Text(
+                    text = "",
+                    style = HmrcTheme.typography.body
+                )
+            }
+        )
+    }
+}
```
jemmaSlater, line 93: This doesn't look like representative usage I don't think.. not sure there's any scenario we would be adding a Text child with no string provided? I think the baseline example would be a primary card with a title but no children...
```suggestion
fun PrimaryCardViewPreviewEmpty() {
    HmrcTheme() {
        PrimaryCardView(title = "Title")
    }
}
```

jemmaSlater (approved):

zertekes07:
PASSED

PR #168, opened 2024-05-14
HMA-7136 ViewExt

emmaatkins:
# 📝 Description

Github Issue
  
- [x] Updated CHANGELOG
- [ ] Updated README

# 🛠 Testing Notes
Second example on SummaryRowVIew content description updated

# 📸 Screenshots

### DIFF
--- a/components/src/main/java/uk/gov/hmrc/components/extensions/ViewExt.kt
+++ b/components/src/main/java/uk/gov/hmrc/components/extensions/ViewExt.kt
@@ -23,7 +23,7 @@ import androidx.core.view.ViewCompat
 import androidx.core.view.accessibility.AccessibilityNodeInfoCompat
 import uk.gov.hmrc.components.R
 
-fun View.setAccessibilityMessage(message: CharSequence) {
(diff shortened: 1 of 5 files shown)


components/src/main/java/uk/gov/hmrc/components/extensions/ViewExt.kt
```diff
@@ -23,7 +23,7 @@ import androidx.core.view.ViewCompat
 import androidx.core.view.accessibility.AccessibilityNodeInfoCompat
 import uk.gov.hmrc.components.R
 
-fun View.setAccessibilityMessage(message: CharSequence) {
+fun View.setAccessibilityMessage(message: CharSequence, className: String?) {
```
jemmaSlater, line 26: Can we add a default parameter here so it's not a breaking change and we don't need to update all our existing usages?
```suggestion
fun View.setAccessibilityMessage(message: CharSequence, className: String? = null) {
```

jemmaSlater (dismissed):

jemmaSlater (approved):

jemmaSlater (approved):

PR #191, opened 2024-09-05
HMA-8827-password-input-view

emmaatkins:
# 📝 Description

HMA-8827
  
- [ ] Updated CHANGELOG
- [ ] Updated README

# 🛠 Testing Notes
Check text input view, currency input view, password input view

# 📸 Screenshots
passwordinputss

### DIFF
--- a/components-compose/src/main/java/uk/gov/hmrc/components/compose/molecule/input/PasswordInputView.kt
+++ b/components-compose/src/main/java/uk/gov/hmrc/components/compose/molecule/input/PasswordInputView.kt
@@ -1,0 +1,108 @@
+/*
+ * Copyright 2023 HM Revenue & Customs
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package uk.gov.hmrc.components.compose.molecule.input
+
+import androidx.compose.foundation.clickable
+import androidx.compose.foundation.text.KeyboardOptions
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.remember
+import androidx.compose.runtime.saveable.rememberSaveable
+import androidx.compose.runtime.setValue
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.text.input.KeyboardType
+import androidx.compose.ui.text.input.PasswordVisualTransformation
+import androidx.compose.ui.text.input.VisualTransformation
+import androidx.compose.ui.tooling.preview.Preview
+import uk.gov.hmrc.components.compose.atom.text.LinkText
+import uk.gov.hmrc.components.compose.ui.theme.HmrcTheme
+
+@Composable
+fun PasswordInputView(
+    modifier: Modifier = Modifier,
+    value: String? = null,
+    onInputValueChange: ((String) -> Unit)? = null,
+    labelText: String? = null,
+    labelContentDescription: String? = null,
+    hintText: String? = null,
+    hintContentDescription: String? = null,
+    placeholderText: String? = null,
+    errorText: String? = null,
+    errorContentDescription: String? = null,
+    singleLine: Boolean = true,
+    numericOnly: Boolean = false,
+    maxChars: Int? = null,
+) {
+
+    // pattern matches a non decimal number
+    val nonDecimalPattern = remember { Regex("^([0-9]*)$") }
+
+    var passwordVisualTransformation: VisualTransformation by remember { mutableStateOf(PasswordVisualTransformation()) }
+
+    var passwordBool: Boolean by rememberSaveable { mutableStateOf(false) }
+
+    fun decimalPatternChecker(input: String, localValue: String) = when {
+        numericOnly && input.matches(nonDecimalPattern) -> input
+        !numericOnly -> input
+        else -> localValue
+    }
+
+    val passwordTrailingIcon: @Composable (() -> Unit) = @Composable {
+        if(passwordBool) {
+           LinkText("Hide", modifier = Modifier.clickable {passwordBool = false})
+            passwordVisualTransformation = VisualTransformation.None
+        }
+        else {
+            LinkText("Show", modifier = Modifier.clickable {passwordBool = true})
+            passwordVisualTransformation = PasswordVisualTransformation()
+        }
+    }
+
+    TextInputView(
+        modifier = modifier,
+        value = value,
+        onInputValueChange = onInputValueChange,
+        inputFilter = { it: String, localValue: String -> decimalPatternChecker(it, localValue) },
+        labelText = labelText,
+        labelContentDescription = labelContentDescription,
+        hintText = hintText,
+        hintContentDescription = hintContentDescription,
+        placeholderText = placeholderText,
+        errorText = errorText,
+        errorContentDescription = errorContentDescription,
+        trailingIcon = passwordTrailingIcon,
+        singleLine = singleLine,
+        keyboardOptions = KeyboardOptions(
+            keyboardType = if (numericOnly) KeyboardType.NumberPassword else KeyboardType.Password
+        ),
+        maxChars = maxChars,
+        visualTransformation = passwordVisualTransformation
+    )
+}
+
+@Preview(showBackground = true)
+@Composable
+fun PasswordInputViewPreview() {
+    HmrcTheme {
+        PasswordInputView(
+            onInputValueChange = { },
+            labelText = "Label",
+            hintText = "Hint",
+            placeholderText = "Text"
+        )
+    }
+}
--- a/components-compose/src/main/java/uk/gov/hmrc/components/compose/molecule/input/TextInputView.kt
+++ b/components-compose/src/main/java/uk/gov/hmrc/components/compose/molecule/input/TextInputView.kt
@@ -37,6 +37,7 @@ import androidx.compose.ui.Modifier
 import androidx.compose.ui.res.stringResource
 import androidx.compose.ui.semantics.contentDescription
 import androidx.compose.ui.semantics.semantics
+import androidx.compose.ui.text.input.VisualTransformation
 import androidx.compose.ui.text.style.TextAlign
 import androidx.compose.ui.tooling.preview.Preview
 import androidx.compose.ui.unit.dp
@@ -64,13 +65,15 @@ object TextInputView {
         hintText: String? = null,
         hintContentDescription: String? = null,
         prefix: @Composable (() -> Unit)? = null,
+        trailingIcon: @Composable (() -> Unit)? = null,
         placeholderText: String? = null,
         errorText: String? = null,
         errorContentDescription: String? = null,
         characterCount: Int? = null,
         maxChars: Int? = null,
         singleLine: Boolean = false,
-        keyboardOptions: KeyboardOptions = KeyboardOptions.Default
+        keyboardOptions: KeyboardOptions = KeyboardOptions.Default,
+        visualTransformation: VisualTransformation = VisualTransformation.None
     ) {
         var localValue: String by rememberSaveable { mutableStateOf(value.orEmpty()) }
         localValue = value.orEmpty()
@@ -119,7 +122,7 @@ object TextInputView {
             }
         }
 
-        val trailingIcon: @Composable (() -> Unit) = @Composable {
+        val clearTrailingIcon: @Composable (() -> Unit) = @Composable {
             if (localValue.isNotEmpty()) {
                 Icon(
                     Icons.Rounded.Close,
@@ -155,9 +158,10 @@ object TextInputView {
                 prefix = prefix,
                 placeholderText = { placeholderText?.let { Text(text = it) } },
                 supportingText = if (counterEnabled) errorTextCounterCombo else error,
-                trailingIcon = trailingIcon,
+                trailingIcon = trailingIcon ?: clearTrailingIcon,
                 singleLine = singleLine,
                 keyboardOptions = keyboardOptions,
+                visualTransformation = visualTransformation
             )
         }
     }
@@ -199,6 +203,7 @@ object TextInputView {
         trailingIcon: @Composable (() -> Unit),
         singleLine: Boolean,
         keyboardOptions: KeyboardOptions,
+        visualTransformation: VisualTransformation
     ) {
         OutlinedTextField(
             modifier = modifier,
@@ -213,7 +218,8 @@ object TextInputView {
             keyboardOptions = keyboardOptions,
             textStyle = typography.body,
             colors = HmrcTheme.textFieldColors,
-            shape = RoundedCornerShape(0)
+            shape = RoundedCornerShape(0),
+            visualTransformation = visualTransformation
         )
     }
 }
--- a/sample-compose-fragments/src/main/java/uk/gov/hmrc/sample_compose_fragments/presentation/fragment/molecules/PasswordInputViewFragment.kt
+++ b/sample-compose-fragments/src/main/java/uk/gov/hmrc/sample_compose_fragments/presentation/fragment/molecules/PasswordInputViewFragment.kt
@@ -1,0 +1,48 @@
+/*
+ * Copyright 2023 HM Revenue & Customs
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package uk.gov.hmrc.sample_compose_fragments.presentation.fragment.molecules
+
+import android.os.Bundle
+import android.view.View
+import androidx.compose.ui.platform.ViewCompositionStrategy
+import androidx.fragment.app.Fragment
+import uk.gov.hmrc.components.compose.ui.theme.HmrcTheme
+import uk.gov.hmrc.components.compose.ui.theme.rememberWindowSizeClass
+import uk.gov.hmrc.sample_compose_components.R
+import uk.gov.hmrc.sample_compose_components.databinding.FragmentComposeExampleBinding
+import uk.gov.hmrc.sample_compose_fragments.presentation.screens.molecules.PasswordInputViewScreen
+import uk.gov.hmrc.sample_compose_fragments.presentation.screens.sampletemplate.HmrcSurface
+
+class PasswordInputViewFragment : Fragment(R.layout.fragment_compose_example) {
+
+    private lateinit var binding: FragmentComposeExampleBinding
+
+    override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
+        super.onViewCreated(view, savedInstanceState)
+        binding = FragmentComposeExampleBinding.bind(view)
+        binding.composeView.apply {
+            setViewCompositionStrategy(ViewCompositionStrategy.DisposeOnViewTreeLifecycleDestroyed)
+            setContent {
+                val window = rememberWindowSizeClass()
+                HmrcTheme(window) {
+                    HmrcSurface {
+                        PasswordInputViewScreen()
+                    }
+                }
+            }
+        }
+    }
+}
--- a/sample-compose-fragments/src/main/java/uk/gov/hmrc/sample_compose_fragments/presentation/screens/molecules/PasswordInputViewScreen.kt
+++ b/sample-compose-fragments/src/main/java/uk/gov/hmrc/sample_compose_fragments/presentation/screens/molecules/PasswordInputViewScreen.kt
@@ -1,0 +1,111 @@
+/*
+ * Copyright 2023 HM Revenue & Customs
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package uk.gov.hmrc.sample_compose_fragments.presentation.screens.molecules
+
+import androidx.compose.foundation.layout.padding
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.saveable.rememberSaveable
+import androidx.compose.runtime.setValue
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.res.stringResource
+import androidx.lifecycle.compose.collectAsStateWithLifecycle
+import androidx.lifecycle.viewmodel.compose.viewModel
+import uk.gov.hmrc.components.compose.molecule.input.PasswordInputView
+import uk.gov.hmrc.components.compose.organism.HmrcCardView
+import uk.gov.hmrc.components.compose.ui.theme.HmrcTheme
+import uk.gov.hmrc.sample_compose_components.R
+import uk.gov.hmrc.sample_compose_fragments.presentation.screens.sampletemplate.ExamplesSlot
+import uk.gov.hmrc.sample_compose_fragments.presentation.screens.sampletemplate.PlaceholderSlot
+import uk.gov.hmrc.sample_compose_fragments.presentation.screens.sampletemplate.ScreenScrollViewColumn
+import uk.gov.hmrc.sample_compose_fragments.presentation.viewModel.PasswordInputViewModel
+
+@Composable
+fun PasswordInputViewScreen() {
+
+    val viewModel = viewModel<PasswordInputViewModel>()
+
+    val errorText = stringResource(id = R.string.password_input_example_error)
+
+    var placeholderValue: String by rememberSaveable { mutableStateOf("") }
+    var example1Value: String by rememberSaveable { mutableStateOf("") }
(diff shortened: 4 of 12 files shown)


components-compose/src/main/java/uk/gov/hmrc/components/compose/molecule/input/PasswordInputView.kt
```diff
@@ -0,0 +1,376 @@
...
+    hintText: String? = null,
+    hintContentDescription: String? = null,
+    placeholderText: String? = null,
+    errorText: String? = null,
+    errorContentDescription: String? = null,
+    showButtonContentDescription: String? = null,
+    hideButtonContentDescription: String? = null,
+    showButtonStateDescription: String? = null,
+    hideButtonStateDescription: String? = null,
+    singleLine: Boolean = true,
```
jemmaSlater, line 85: Will it break the design if the consumer sets this to false? I think we can remove this as a parameter and have it enforced within this component

components-compose/src/main/java/uk/gov/hmrc/components/compose/molecule/input/PasswordInputView.kt
```diff
@@ -0,0 +1,376 @@
...
+        animateDpAsState(targetThickness, tween(durationMillis = ANIMATION_DURATION))
+    } else {
+        rememberUpdatedState(unfocusedBorderThickness)
+    }
+    return rememberUpdatedState(
+        BorderStroke(animatedThickness.value, SolidColor(indicatorColor.value))
+    )
+}
+
+var borderStroke: State<BorderStroke>? = null
```
jemmaSlater, line 154: Is this meant to be a global var? Should it live inside one of the functions?

components-compose/src/main/java/uk/gov/hmrc/components/compose/molecule/input/PasswordInputView.kt
```diff
@@ -0,0 +1,376 @@
...
+
+@Composable
+fun CalculateBorder(
+    enabled: Boolean = true,
```
jemmaSlater, line 158: This appears to always be true. Do we need to provide it as a parameter at all?

components-compose/src/main/java/uk/gov/hmrc/components/compose/molecule/input/PasswordInputView.kt
```diff
@@ -0,0 +1,376 @@
...
+    }
+    return rememberUpdatedState(
+        BorderStroke(animatedThickness.value, SolidColor(indicatorColor.value))
+    )
+}
+
+var borderStroke: State<BorderStroke>? = null
+
+@Composable
+fun CalculateBorder(
```
jemmaSlater, line 157: I think we should make this private so that it's only able to be used within this class
```suggestion
private fun CalculateBorder(
```

components-compose/src/main/java/uk/gov/hmrc/components/compose/molecule/input/PasswordInputView.kt
```diff
@@ -0,0 +1,376 @@
...
+            keyboardType = if (numericOnly) KeyboardType.NumberPassword else KeyboardType.Password
+        ),
+        interactionSource = interactionSource,
+        requiredSequencesSpacing = requiredSequencesSpacing,
+        textStyle = textStyle
+    )
+}
+
+@Composable
+fun animateBorderStrokeAsState(
```
jemmaSlater, line 134:
```suggestion
private fun animateBorderStrokeAsState(
```

components-compose/src/main/java/uk/gov/hmrc/components/compose/molecule/input/PasswordInputView.kt
```diff
@@ -0,0 +1,376 @@
...
+
+var borderStroke: State<BorderStroke>? = null
+
+@Composable
+fun CalculateBorder(
+    enabled: Boolean = true,
+    isError: Boolean,
+    interactionSource: InteractionSource,
+    focusedBorderThickness: Dp = FocusedBorderThickness,
+    unfocusedBorderThickness: Dp = UnfocusedBorderThickness
```
jemmaSlater, line 162: I don't think we need to declare these as parameters since we never need to change them. We can use `FocusedBorderThickness` and `UnfocusedBorderThickness` directly inside the `animateBorderStrokeAsState` function, and save a lot of code repetition in the parameters.

components-compose/src/main/java/uk/gov/hmrc/components/compose/molecule/input/PasswordInputView.kt
```diff
@@ -0,0 +1,376 @@
...
+) {
+    borderStroke = animateBorderStrokeAsState(
+        enabled,
+        isError,
+        interactionSource,
+        focusedBorderThickness,
+        unfocusedBorderThickness
+    )
```
jemmaSlater, line 170: Is this function needed? Could we call `animateBorderStrokeAsState` directly instead and reduce the code?

components-compose/src/main/java/uk/gov/hmrc/components/compose/molecule/input/PasswordInputView.kt
```diff
@@ -0,0 +1,376 @@
...
+
+@Composable
+fun indicatorColor(
+    enabled: Boolean,
+    isError: Boolean,
+    interactionSource: InteractionSource
+): State<Color> {
+    val focused by interactionSource.collectIsFocusedAsState()
+
+    val colors = HmrcTheme.colors
```
jemmaSlater, line 181: Have you checked this works when toggling between light and dark mode?

components-compose/src/main/java/uk/gov/hmrc/components/compose/molecule/input/PasswordInputView.kt
```diff
@@ -0,0 +1,376 @@
...
+    labelText: String? = null,
+    labelContentDescription: String? = null,
+    hintText: String? = null,
+    hintContentDescription: String? = null,
+    prefix: @Composable() (() -> Unit)? = null,
+    placeholderText: String? = null,
+    errorText: String? = null,
+    errorContentDescription: String? = null,
+    showButtonText: String? = stringResource(R.string.passwordTextInputView_show),
+    hideButtonText: String? = stringResource(R.string.passwordTextInputView_hide),
```
jemmaSlater, line 215: I think it would be safer to not provide default parameters here, as we want the consumer to provide localised versions of the strings.

components-compose/src/main/java/uk/gov/hmrc/components/compose/molecule/input/PasswordInputView.kt
```diff
@@ -0,0 +1,376 @@
...
+        enabled,
+        isError,
+        interactionSource,
+        focusedBorderThickness,
+        unfocusedBorderThickness
+    )
+}
+
+@Composable
+fun indicatorColor(
```
jemmaSlater, line 174:
```suggestion
private fun indicatorColor(
```

components-compose/src/main/java/uk/gov/hmrc/components/compose/molecule/input/PasswordInputView.kt
```diff
@@ -0,0 +1,376 @@
...
+    return if (enabled) {
+        animateColorAsState(targetValue, tween(durationMillis = ANIMATION_DURATION))
+    } else {
+        rememberUpdatedState(targetValue)
+    }
+}
+
+@SuppressWarnings("LongMethod", "ComplexMethod")
+@Composable
+fun PasswordTextInputView(
```
jemmaSlater, line 201:
```suggestion
private fun PasswordTextInputView(
```

components-compose/src/main/java/uk/gov/hmrc/components/compose/molecule/input/PasswordInputView.kt
```diff
@@ -0,0 +1,376 @@
...
+    modifier: Modifier = Modifier,
+    value: String? = null,
+    onInputValueChange: ((String) -> Unit)? = null,
+    inputFilter: ((String, String) -> String)? = null,
+    labelText: String? = null,
+    labelContentDescription: String? = null,
+    hintText: String? = null,
+    hintContentDescription: String? = null,
+    prefix: @Composable() (() -> Unit)? = null,
+    placeholderText: String? = null,
```
jemmaSlater, line 211: Will the password field ever have a prefix or placeholder text? I think we can remove these

components-compose/src/main/java/uk/gov/hmrc/components/compose/molecule/input/PasswordInputView.kt
```diff
@@ -0,0 +1,376 @@
...
+    hideButtonContentDescription: String? = null,
+    showButtonStateDescription: String? = null,
+    hideButtonStateDescription: String? = null,
+    characterCount: Int? = null,
+    maxChars: Int? = null,
+    singleLine: Boolean = false,
+    keyboardOptions: KeyboardOptions = KeyboardOptions.Default,
+    interactionSource: MutableInteractionSource,
+    requiredSequencesSpacing: Boolean,
+    textStyle: TextStyle
```
jemmaSlater, line 226: Same comments as above for these parameters.. do we need them all? Some of them we can hardcode as we don't want the user to be able to change them. Although this function isn't exposed to the user, we can save code by defining our set rules at the earliest point, which is inside the `TextField` declaration in this function.

components-compose/src/main/java/uk/gov/hmrc/components/compose/molecule/input/PasswordInputView.kt
```diff
@@ -0,0 +1,376 @@
...
+
+    isError = !localError.isNullOrEmpty() || (localValue.length > (characterCount ?: Int.MAX_VALUE))
+
+    @Composable
+    fun ShowHideButton(
+        modifier: Modifier,
+        hideButtonContentDescription: String?,
+        hideButtonStateDescription: String?,
+        showButtonContentDescription: String?,
+        showButtonStateDescription: String?,
```
jemmaSlater, line 246: Do we actually want these to be optional?

components-compose/src/main/java/uk/gov/hmrc/components/compose/molecule/input/PasswordInputView.kt
```diff
@@ -0,0 +1,376 @@
...
+    var localError: String? by rememberSaveable { mutableStateOf(null) }
+    localError = errorText
+
+    var isError: Boolean by rememberSaveable { mutableStateOf(false) }
+
+    isError = !localError.isNullOrEmpty() || (localValue.length > (characterCount ?: Int.MAX_VALUE))
+
+    @Composable
+    fun ShowHideButton(
+        modifier: Modifier,
```
jemmaSlater, line 242: We don't need this modifier parameter as we never pass anything through here to it.

components-compose/src/main/java/uk/gov/hmrc/components/compose/molecule/input/PasswordInputView.kt
```diff
@@ -0,0 +1,376 @@
...
+        isError,
+        interactionSource,
+        focusedBorderThickness = FocusedBorderThickness,
+        unfocusedBorderThickness = UnfocusedBorderThickness
+    )
+    Column(modifier = modifier) {
+        Label(labelText = labelText, labelContentDescription = labelContentDescription)
+        Hint(hintText = hintText, hintContentDescription = hintContentDescription)
+        Row(
+            borderStroke?.let {
```
jemmaSlater, line 282: Should this be nullable? Won't we always have a value from `animateBorderStrokeAsState` which doesn't have a null state. I think we might be able to simplify this.

components-compose/src/main/java/uk/gov/hmrc/components/compose/molecule/input/PasswordInputView.kt
```diff
@@ -0,0 +1,376 @@
...
+                    .border(it.value, RectangleShape)
+            } ?: Modifier,
+            verticalAlignment = Alignment.CenterVertically
+        ) {
+            Column(Modifier.weight(1f)) {
+                TextField(
+                    modifier = if (isError) Modifier.clearAndSetSemantics {
+                        localError?.let {
+                            error(
+                                "$it. $errorContentDescription"
```
jemmaSlater, line 294: Will this duplicate the error announcement? Will it read eg. "too many characters. Error too many characters"? I think `localError` is the visible error text?

components-compose/src/main/java/uk/gov/hmrc/components/compose/molecule/input/PasswordInputView.kt
```diff
@@ -0,0 +1,376 @@
...
+                            localValue = if (inputFilter != null && newValue.isNotEmpty()) {
+                                inputFilter(newValue, localValue)
+                            } else newValue
+                            if (onInputValueChange != null) {
+                                onInputValueChange(localValue)
+                            }
+                        }
+                    },
+                    prefix = prefix,
+                    placeholderText = { placeholderText?.let { Text(text = it) } },
```
jemmaSlater, line 311: I think we can remove this because we don't want placeholder text on this component.

components-compose/src/main/java/uk/gov/hmrc/components/compose/molecule/input/PasswordInputView.kt
```diff
@@ -0,0 +1,376 @@
...
+
+            Column(Modifier.width(IntrinsicSize.Max)) {
+                ShowHideButton(
+                    Modifier,
+                    hideButtonContentDescription,
+                    hideButtonStateDescription,
+                    showButtonContentDescription,
+                    showButtonStateDescription,
+                    showButtonText ?: stringResource(R.string.passwordTextInputView_show),
+                    hideButtonText ?: stringResource(R.string.passwordTextInputView_hide)
```
jemmaSlater, line 343: I don't think we want a default here. I think `showButtonText` and `hideButtonText` should be required and then it won't be possible to have a null state.

components-compose/src/main/java/uk/gov/hmrc/components/compose/molecule/input/TextInputView.kt
```diff
@@ -64,13 +50,18 @@ object TextInputView {
         hintText: String? = null,
         hintContentDescription: String? = null,
         prefix: @Composable (() -> Unit)? = null,
+        trailingIcon: @Composable (() -> Unit)? = null,
```
jemmaSlater, line 53: Do we need to add this?

components-compose/src/main/java/uk/gov/hmrc/components/compose/molecule/input/TextInputView.kt
```diff
@@ -64,13 +50,18 @@ object TextInputView {
...
         placeholderText: String? = null,
         errorText: String? = null,
         errorContentDescription: String? = null,
         characterCount: Int? = null,
         maxChars: Int? = null,
         singleLine: Boolean = false,
-        keyboardOptions: KeyboardOptions = KeyboardOptions.Default
+        keyboardOptions: KeyboardOptions = KeyboardOptions.Default,
+        visualTransformation: VisualTransformation = VisualTransformation.None,
+        interactionSource: MutableInteractionSource = remember { MutableInteractionSource() },
```
jemmaSlater, line 62: Do we need to add these?

components-compose/src/main/java/uk/gov/hmrc/components/compose/molecule/input/TextInputView.kt
```diff
@@ -80,68 +71,26 @@ object TextInputView {
...
-                        textAlign = TextAlign.End,
-                    )
-                }
-            }
-        }
-
-        val trailingIcon: @Composable (() -> Unit) = @Composable {
-            if (localValue.isNotEmpty()) {
+        val clearTrailingIcon: @Composable (() -> Unit) = @Composable {
+            if (false) {
```
jemmaSlater, line 75: I'm not sure what this is doing... perhaps a leftover experiment? :)

components-compose/src/main/java/uk/gov/hmrc/components/compose/molecule/input/TextInputView.kt
```diff
@@ -154,68 +103,21 @@ object TextInputView {
...
+                else error(
+                    errorText,
+                    errorContentDescription
+                ),
                 singleLine = singleLine,
                 keyboardOptions = keyboardOptions,
+                visualTransformation = visualTransformation,
+                trailingIcon = trailingIcon ?: clearTrailingIcon,
+                colors = HmrcTheme.textFieldColors,
+                interactionSource = interactionSource
```
jemmaSlater, line 117: I don't think we need all these? visual transformation, colors, interaction source don't need to be defined I think... not sure about trailing icon - we don't want to change whatever the existing behaviour was with the clear button.

components-compose/src/main/java/uk/gov/hmrc/components/compose/ui/theme/Color.kt
```diff
@@ -33,6 +33,7 @@ val HmrcPink = Color(0xFFCA2B75)
 val HmrcYellow = Color(0xFFFFBF47)
 val HmrcAlwaysBlack = Color(0xFF0B0C0C)
 val HmrcDonutBlue = Color(0xFF003078)
+val hmrcTransparent = Color.Transparent
```
jemmaSlater, line 36:
```suggestion
val HmrcTransparent = Color.Transparent
```

components-compose/src/main/res/values/strings.xml
```diff
@@ -19,4 +19,7 @@
     <string name="accessibility_warning">warning; %s</string>
     <string name="bottomSheetView_drag_handle_content_desc">Drag handle</string>
     <string name="bottomSheetView_close_content_desc">Dismiss bottom sheet</string>
+    <string name="passwordTextInputView_hide">Hide</string>
+    <string name="passwordTextInputView_show">Show</string>
```
jemmaSlater, line 23: I don't think we want to include these in the actual component library package, but they can be declared in the sample package so we can use them in the examples.

sample-compose-fragments/src/main/java/uk/gov/hmrc/sample_compose_fragments/presentation/fragment/molecules/PasswordInputViewFragment.kt
```diff
@@ -0,0 +1,48 @@
+/*
+ * Copyright 2023 HM Revenue & Customs
```
jemmaSlater, line 2:
```suggestion
 * Copyright 2024 HM Revenue & Customs
```

sample-compose-fragments/src/main/java/uk/gov/hmrc/sample_compose_fragments/presentation/screens/molecules/PasswordInputViewScreen.kt
```diff
@@ -0,0 +1,114 @@
...
+    var example3Value: String by rememberSaveable { mutableStateOf("Password123") }
+
+    ScreenScrollViewColumn {
+        PlaceholderSlot {
+            PasswordInputView(
+                value = placeholderValue,
+                onInputValueChange = { placeholderValue = it },
+                labelText = stringResource(id = R.string.password_input_placeholder_label),
+                hintText = stringResource(id = R.string.password_input_placeholder_hint),
+                placeholderText = stringResource(id = R.string.password_input_placeholder_placeholder),
```
jemmaSlater, line 56: This placeholder text isn't needed

sample-compose-fragments/src/main/java/uk/gov/hmrc/sample_compose_fragments/presentation/screens/molecules/PasswordInputViewScreen.kt
```diff
@@ -0,0 +1,114 @@
...
+                    value = example3Value,
+                    onInputValueChange = {
+                        viewModel.isEmptyValidation(it, errorText, 2)
+                        example3Value = it
+                    },
+                    errorText = viewModel.textInputErrorEmptyValidation2.collectAsStateWithLifecycle().value,
+                    labelText = stringResource(id = R.string.password_input_example_3_label),
+                    hintText = stringResource(id = R.string.password_input_example_3_hint),
+                )
+            }
```
jemmaSlater, line 111: Could we add an example that showcases all the accessibility stuff so it can be tested, like adding different content descriptions and button text?

components-compose/src/main/java/uk/gov/hmrc/components/compose/molecule/input/PasswordInputView.kt
emmaatkins: Yes :) It works perfectly

sample-compose-fragments/src/main/java/uk/gov/hmrc/sample_compose_fragments/presentation/screens/molecules/PasswordInputViewScreen.kt
emmaatkins: Put it all on ex1 now

components-compose/src/main/java/uk/gov/hmrc/components/compose/molecule/input/PasswordInputView.kt
emmaatkins: Yes - we don't need them and there's fallbacks if user doesn't supply them

components-compose/src/main/java/uk/gov/hmrc/components/compose/molecule/input/TextInputView.kt
emmaatkins: colors was already defined as such, trailingicon I just renamed to make more clear (pun intended!)

components-compose/src/main/java/uk/gov/hmrc/components/compose/molecule/input/PasswordInputView.kt
```diff
@@ -0,0 +1,343 @@
...
+        errorText = errorText,
+        errorContentDescription = errorContentDescription,
+        passwordTrailingButton = PasswordTrailingButton(
+            showButtonText = passwordTrailingButton.showButtonText,
+            showButtonContentDescription = passwordTrailingButton.showButtonContentDescription,
+            showButtonStateDescription = passwordTrailingButton.showButtonStateDescription,
+            hideButtonText = passwordTrailingButton.hideButtonText,
+            hideButtonContentDescription = passwordTrailingButton.hideButtonContentDescription,
+            hideButtonStateDescription = passwordTrailingButton.hideButtonStateDescription
+        ),
```
jemmaSlater, line 117: Do we need to do this? Could we just pass the `passwordTrailingButton` through? Looks like they're the same object type already, so not sure we need to map it to a new object?

components-compose/src/main/java/uk/gov/hmrc/components/compose/molecule/input/PasswordInputView.kt
jemmaSlater: I don't think we want to expose this part to the consumer so should be private :)

components-compose/src/main/java/uk/gov/hmrc/components/compose/molecule/input/PasswordInputView.kt
```diff
@@ -0,0 +1,343 @@
...
+
+        if (hideButtonShown) {
+            SecondaryButton(
+                hideButtonText, onClick = { hideButtonShown = false },
+                modifier = Modifier
+                    .clearAndSetSemantics {
+                        contentDescription = hideButtonContentDescription ?: hideButtonText
+                        hideButtonStateDescription?.let { stateDescription = it }
+                        role = Role.Button
+                        onClick(actionString, action = { hideButtonShown = false; true })
```
jemmaSlater, line 211: Do we need the `true` here? What does it do?

components-compose/src/main/java/uk/gov/hmrc/components/compose/molecule/input/PasswordInputView.kt
```diff
@@ -0,0 +1,343 @@
...
+            )
+        } else {
+            SecondaryButton(
+                showButtonText, onClick = { hideButtonShown = true },
+                modifier = Modifier
+                    .clearAndSetSemantics {
+                        contentDescription = showButtonContentDescription ?: showButtonText
+                        showButtonStateDescription?.let { stateDescription = it }
+                        role = Role.Button
+                        onClick(actionString, action = { hideButtonShown = true; true })
```
jemmaSlater, line 222: Do we need the true here? What does it do?

components-compose/src/main/java/uk/gov/hmrc/components/compose/molecule/input/PasswordInputView.kt
```diff
@@ -0,0 +1,343 @@
...
+    )
+    Column(modifier = modifier) {
+        Label(labelText = labelText, labelContentDescription = labelContentDescription)
+        Hint(hintText = hintText, hintContentDescription = hintContentDescription)
+        Row(
+            borderStroke.let {
+                Modifier
+                    .fillMaxWidth()
+                    .border(it.value, RectangleShape)
+            },
```
jemmaSlater, line 240:
```suggestion
            Modifier
                    .fillMaxWidth()
                    .border(borderStroke.value, RectangleShape)
            },
```

components-compose/src/main/java/uk/gov/hmrc/components/compose/molecule/input/TextInputView.kt
```diff
@@ -63,14 +47,17 @@ object TextInputView {
...
         placeholderText: String? = null,
         errorText: String? = null,
         errorContentDescription: String? = null,
         characterCount: Int? = null,
         maxChars: Int? = null,
         singleLine: Boolean = false,
-        keyboardOptions: KeyboardOptions = KeyboardOptions.Default
+        keyboardOptions: KeyboardOptions = KeyboardOptions.Default,
+        visualTransformation: VisualTransformation = VisualTransformation.None,
+        requiredSequencesSpacing: Boolean = false
```
jemmaSlater, line 59: It doesn't look like this is actually used anywhere currently to achieve the spacing for sequences..

components-compose/src/main/java/uk/gov/hmrc/components/compose/molecule/input/TextInputView.kt
```diff
@@ -63,14 +47,17 @@ object TextInputView {
...
+        prefix: @Composable() (() -> Unit)? = null,
         placeholderText: String? = null,
         errorText: String? = null,
         errorContentDescription: String? = null,
         characterCount: Int? = null,
         maxChars: Int? = null,
         singleLine: Boolean = false,
-        keyboardOptions: KeyboardOptions = KeyboardOptions.Default
+        keyboardOptions: KeyboardOptions = KeyboardOptions.Default,
+        visualTransformation: VisualTransformation = VisualTransformation.None,
```
jemmaSlater, line 58: Do we want people to be able to change the visual transformation for the normal text input view?

components-compose/src/main/java/uk/gov/hmrc/components/compose/molecule/input/PasswordInputView.kt
emmaatkins: It doesn't work without it and I'm following the same code layout as the Google codelabs - I don't understand the exact specifics though

jemmaSlater (approved):

jemmaSlater (approved):

PR #192, opened 2024-09-30
Update modifier usage for buttons

jemmaSlater:
# 📝 Description

The previous implementation meant that if, for example, a modifier with some padding was added to the implementation of a button, it would mess up the spacing within the button elements such as text or icons as they were using the same modifier. This has now been fixed.
  
- [x] Updated CHANGELOG
- [ ] Updated README

# 🛠 Testing Notes

# 📸 Screenshots

### DIFF
--- a/components-compose/src/main/java/uk/gov/hmrc/components/compose/atom/button/Button.kt
+++ b/components-compose/src/main/java/uk/gov/hmrc/components/compose/atom/button/Button.kt
@@ -72,7 +72,7 @@ fun HmrcButton(
             fontSize = 16.sp,
             fontWeight = FontWeight.Normal,
             textAlign = textAlign,
-            modifier = modifier.weight(1f)
+            modifier = Modifier.weight(1f)
         )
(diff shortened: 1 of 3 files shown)


components-compose/src/main/java/uk/gov/hmrc/components/compose/atom/button/Button.kt
```diff
@@ -134,12 +134,12 @@ fun IconButton(
     textAlign: TextAlign = TextAlign.Start,
     onClick: () -> Unit,
 ) {
-    SecondaryButton(text = text, modifier = Modifier, textAlign = textAlign, onClick = onClick) {
+    SecondaryButton(text = text, modifier = modifier, textAlign = textAlign, onClick = onClick) {
```
aliahmedbd, line 137: Jemma, I can't find any usage of this modifier -> "modifier"

annadahmrc (approved):

PR #220, opened 2026-06-17
HMA-8834: Upgrade Ruby to latest for release

james-crook:


rajesh-hmrc:
Not required anymore as versions are already updated

